Issue #10, opened 2018-07-08
The tests aren't cleaning up after themselves


Skarlso:
When I run `go test` a bunch of things are left under `pipeline` that pollute the work tree.

```bash
~/golang/src/github.com/gaia-pipeline/gaia makefile_refactor* 10s
❯ git status
On branch makefile_refactor
Your branch is up-to-date with 'skarlso/makefile_refactor'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	pipeline/.circleci/
	pipeline/.gitignore
	pipeline/.goreleaser.yml
	pipeline/Gopkg.lock
	pipeline/Gopkg.toml
	pipeline/LICENSE
	pipeline/Makefile
	pipeline/README.rst
	pipeline/cmd/
	pipeline/docker/
	pipeline/frontend/
	pipeline/gaia.go
	pipeline/handlers/
	pipeline/pipeline/
	pipeline/plugin/
	pipeline/scheduler/
	pipeline/store/
	pipeline/vendor/
```

michelvocks:
I'm actually a bit ashamed of the tests. 😣
They are really messed up and most times they are not working.

Skarlso:
Don't worry about it. :) I'll clean them up as I go through them. ;)

Skarlso:
Fixed by #12

PR #12, opened 2018-07-08
Fixed git checkout to define a tmp directory

Skarlso:


pipeline/git_test.go
```diff
@@ -9,16 +9,13 @@ import (
 
 func TestGitCloneRepo(t *testing.T) {
 	repo := &gaia.GitRepo{
-		URL: "https://github.com/gaia-pipeline/gaia",
+		URL:       "https://github.com/gaia-pipeline/gaia",
```
Skarlso: Consider creating a test repo under `gaia-pipeline` so that it checks out that instead of the entire gaia repo which will potentially be large at some point. Also, this test right now takes about 8 seconds which isn't very nice. Create a test repo with a README in it in order to verify this functionality.

Also, `LocalDest` wasn't defined and `RemoveAll` wasn't called with defer, which would end up as being not cleaned if `gitCloneRepo` would have failed.

Skarlso:
Fixes #10

PR #27, opened 2018-07-12
use rsa key for jwt

### LINKED ISSUE #26: JWT not valid afteer restart
Restarting the gaia process invalidates all JWT tokens. It seems like the jwt key is auto generated on startup instead of being passed in via a flag.

rmb938:
Fixes #26

### DIFF
--- a/handlers/handler.go
+++ b/handlers/handler.go
@@ -1,13 +1,14 @@
 package handlers
 
 import (
-	"crypto/rand"
 	"errors"
 	"fmt"
 	"net/http"
 	"strings"
 
 	"github.com/GeertJohan/go.rice"
+
+	"crypto/rsa"
 
 	jwt "github.com/dgrijalva/jwt-go"
 	"github.com/gaia-pipeline/gaia"
@@ -48,21 +49,11 @@ var storeService *store.Store
 
 var schedulerService *scheduler.Scheduler
 
-// jwtKey is a random generated key for jwt signing
-var jwtKey []byte
-
 // InitHandlers initializes(registers) all handlers
 func InitHandlers(e *echo.Echo, store *store.Store, scheduler *scheduler.Scheduler) error {
 	// Set instances
 	storeService = store
 	schedulerService = scheduler
-
-	// Generate signing key for jwt
-	jwtKey = make([]byte, 64)
-	_, err := rand.Read(jwtKey)
-	if err != nil {
-		return err
-	}
 
 	// Define prefix
 	p := "/api/" + apiVersion + "/"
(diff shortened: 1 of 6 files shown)


Skarlso:
So, I advice against this because... :)

First off, your key would be visible in your bash history even if it's a dev environment that would be a bad idea. Unless you would like to use a vault and pass in an environment property which is a hassle.

Second, the signing should happen via a public / private key-pair to be honest rather than a passed in OR generated jws key. So instead of doing this, try focusing on a key exchange solution?

rmb938:
I can also add an environment variable for the key. 

Also why not support both a key and a public/private key-pair? If someone wants to spin up gaia just to try out or quickly dev with having to first generate a keypair is kinda annoying.

Skarlso:
Because providing the ability to circumvent security because of convenience is just asking for people to use it in production and they definitely will because it's easier. :)

Instead do this: 

* Load in a private key ( for which the name could be passed in from command line )
* And use the content as a security key
* Use the public part of that key as a verify key

Shouldn't be too difficult and would remove the need for yet another flag.

rmb938:
Ok, no problem. I can make those changes later today.

Skarlso:
```go
...
privateKey, _ := ioutil.ReadFile(privateKeyPath)
signKey, _ = jwt.ParseRSAPrivateKeyFromPEM(privateKey)
...
tokenString, err := t.SignedString(signKey)
...
```

Something like this should work juuuust nicely. :)

rmb938:
Yup, that's what I was planning.

Skarlso:
Cool. And here https://github.com/gaia-pipeline/gaia/blob/master/handlers/handler.go#L144 you can use the opposite ParseRSAPublicKeyFromPEM.

Alright, I shuddup. It's just an interesting PR. :)

michelvocks:
Thanks a lot @rmb938 for your contribution. 🤗 

I personally think that the initial startup of Gaia should be easy and simple. 
If I understand your code correctly then currently the user is forced to set the parameter `jwtkey` and like you already said: This will be a bad first experience for all new users.
I also totally agree with @Skarlso point that this is an security issue. 

So what do you think about providing a parameter to set a public/private key and if this parameter is empty we generate a token (like we do it now). Do you think this is still an security issue? :wink:

rmb938:
Yup that makes sense. What do you think about printing a warning message if the auto generated key is being used? That way users know they messed up something if they actually wanted to use a private key.

Skarlso:
@rmb938 @michelvocks Sounds cool! 👍

michelvocks:
Awesome! 🤗 
Looking forward to it!

rmb938:
@michelvocks How does that look?

handlers/handler.go
```diff
@@ -142,13 +133,21 @@ func authBarrier(next echo.HandlerFunc) echo.HandlerFunc {
...
+			case *jwt.SigningMethodHMAC:
+				if _, ok := gaia.Cfg.JWTKey.([]byte); !ok {
+					return nil, signingMethodError
+				}
+				return gaia.Cfg.JWTKey, nil
+			case *jwt.SigningMethodRSA:
+				if _, ok := gaia.Cfg.JWTKey.(*rsa.PrivateKey); !ok {
+					return nil, signingMethodError
+				}
+				return gaia.Cfg.JWTKey.(*rsa.PrivateKey).Public(), nil
```
Skarlso: Oh cool, you can do .Public. Nice.

Skarlso:
This is looking good. :) You'll have to write tests though. ;) :)

rmb938:
@Skarlso @michelvocks Added tests for everything I added so we should be good to go :)

Skarlso:
Looking good! Good job with the tests! 👍  :)

michelvocks:
@rmb938 can you have a short look at my review comment? :relaxed:

Would be nice if you could change this one simple thing.

PR #40, opened 2018-07-18
Add tests to PipelineGitLSRemote

azbshiri:
Tests that check whether PipelineGitLSRemote works properly.

- [x] Invalid Data
- [x] Invalid Access
- [x] Success

### DIFF
--- a/handlers/pipeline_test.go
+++ b/handlers/pipeline_test.go
@@ -13,7 +13,6 @@ import (
 	"github.com/gaia-pipeline/gaia/store"
 	hclog "github.com/hashicorp/go-hclog"
 	"github.com/labstack/echo"
-	"github.com/stretchr/testify/assert"
 )
 
 func TestPipelineGitLSRemote(t *testing.T) {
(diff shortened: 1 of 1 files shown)


azbshiri:
@michelvocks are you ok with this type of tests?

handlers/pipeline_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"testing"
+
+	"github.com/gaia-pipeline/gaia"
+	"github.com/gaia-pipeline/gaia/store"
+	hclog "github.com/hashicorp/go-hclog"
+	"github.com/labstack/echo"
+	"github.com/stretchr/testify/assert"
```
michelvocks: It would be nice if you could remove this dependency. I know assert from other projects (docker for example) and they had at the end huge problems with it. It seems like it's not 100% reliable. :confused:

I know it's more overhead and makes the tests a bit more ugly but at least we are 100% using the standard library which I would like to stick if it's somehow possible. 🤗

azbshiri: done

michelvocks:
LGTM. Thanks a lot 🤗 ❤️

PR #48, opened 2018-07-20
Automatically update pipelines webhook

Skarlso:
This currently a working end-point for accepting and parsing the webhooks. I'm opening a PR for conversation purposes. I'm debating whether to make the user create a webhook or have that authentication process in Gaia complete with callback. 

I'm all for making it as easy as possible for a user and probably a user would say the same thing. Which means Gaia will have to redirect and ask for access to the repository in order to send a create webhook request to Github.

Thoughts?


Skarlso:
@michelvocks Okay, so I went with the easiest solution here and the least possible impact in either the workflow or the user experience. I'm using a github token which the user needs to provide for now as an environment property later on from a secret file of some sort or maybe both.

Right now everything is hardcoded as I was testing if the authentication and the creation is working fine. Here you can see that the hook is created for said repository:

screen shot 2018-07-21 at 09 40 27

This also means that I have to provide a callback url. Which means I will need to be able to determine where and under what address gaia is currently deployed at. Which I'm sure I can. :) 

Also pinging @rmb938 for the glory of the Sontaran Empire.

sontaran

And to see if you are satisfied with the approach laid out. :)

Makefile
```diff
@@ -4,7 +4,7 @@ GO_LDFLAGS_STATIC=-ldflags "-s -w -extldflags -static"
 default: dev
 
 dev:
-	go run ./cmd/gaia/main.go -homepath=${PWD}/tmp -dev=true -poll=true
```
Skarlso: I haven't noticed that I put this in. 😊

handlers/hook.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		e := struct {
+			Message string
+		}{
+			Message: err.Error(),
+		}
+		return c.JSON(400, e)
+	}
+
+	gaia.Cfg.Logger.Info("received: ", h.Event)
```
Skarlso: Actually, I don't care about the event at all. And in any other case, I'm thinking it's only enough to verify the signature. Since the hook is only for one event, this data here is meaningless. The hook will only ever be called on a push. If there is an event, we build the specific pipeline.

pipeline/create_pipeline.go
```diff
@@ -99,4 +99,11 @@ func CreatePipeline(p *gaia.CreatePipeline) {
 		gaia.Cfg.Logger.Error("cannot put create pipeline into store", "error", err.Error())
 		return
 	}
+
+	if !gaia.Cfg.Poll {
+		// TODO: Add repo VCS type information?
+		// TODO: Q: Also, save the webhook so we can later find the pipeline associated with it?
+		// A: Don't think so, because the hook will contain the name of the repository.
+		subscribeRepoWebhook(&p.Pipeline.Repo)
```
Skarlso: @michelvocks This also means that webhook automatic building will never be available for drop-in pipelines as this needs repository information. Are you okay with this? Or shall we try and do some workaround magic?

pipeline/git.go
```diff
@@ -152,3 +157,32 @@ func updateAllCurrentPipelines() {
...
+
+func subscribeRepoWebhook(repo *gaia.GitRepo) error {
+	token := os.Getenv("GAIA_GIT_TOKEN")
+	ctx := context.Background()
+	ts := oauth2.StaticTokenSource(
+		&oauth2.Token{AccessToken: token},
+	)
+	tc := oauth2.NewClient(ctx, ts)
+	config := make(map[string]interface{})
+	config["url"] = "https://example.com/callback"
```
Skarlso: So, unless echo is running and a request was received I can't determine where gaia is deployed at from the back-end side. Maybe the admin site can't, I'm not big on NPM. However, that leaves me with config options maybe. :/ Which further convolutes the huge configuration flags by adding `-host=example.com` or something like that.

I see the config refactoring approaching ever faster. :) What say you?

Makefile
michelvocks: I guess you can leave that in. Especially when you do pipeline testing this is really helpful! 😄

handlers/hook.go
michelvocks: Right. We just have to make sure that this event triggers the right pipeline 😃

pipeline/create_pipeline.go
michelvocks: No, this is absolutely fine! If you just drop your pipeline binary into the pipeline folder, this pipeline cannot use the poll functionality as well as the webhook. 👍

pipeline/git.go
michelvocks: Good point. I know from many tools (gitlab for example) that you have to provide the domain name where the application is accessible from. So I guess adding a configuration flag is fine here. 

I actually think we should do both. The admin can provide the domain name (which will be used for sure) and if it's not provided we try to determine it after the first request came in. What do you think?

Skarlso:
@rmb938 @michelvocks Thanks guys. This is all very good feedback and I agree with the both of you. :)

michelvocks:
@Skarlso Yes sure. Just let me know where I can help explicitly.

Skarlso:
So, what I'm going to do in this Pr is the following... @michelvocks @rmb938.

* Not going to save the git token as there is, for now, no reason for it. It's only used once, and that's when the pipeline is created and the hook is sent.
* Modify the UI to have a checkbox that the user want's a webhook, and if it's checked, popup a field where a github TOKEN is requested. ( make apparent that it's not saved at this time ).
* This means deleting a hook is not handled currently ( that will be a future after we save the token in the form of issue #51 (handling secrets via a vault).
* Modify the auto-updater to be opt-in and save a flag for a pipeline saying (autoupdate - true) because maybe you don't always want all your pipeline to update. Thoughts on this?
* Add a flag called `hostname` to use for the callback url.

I think that's all.

pipeline/git.go
Skarlso: I agree that admin can provide a hostname which overwrites the one provided by the flag. I guess we can try. The first request would be a user login?

Skarlso:
So this works right now:

```bash
2018/07/22 23:08:47 hook created:  "web" 201 Created
2018/07/22 23:08:47 hook url:  https://api.github.com/repos/Skarlso/go-example/hooks/39614928
```

button:

screen shot 2018-07-22 at 23 00 43
screen shot 2018-07-22 at 23 00 36

The only two problems are that the token is saved if the pipeline errors out and second, I still have no way of providing and saving a `secret` which brings us back to #51. :/

@michelvocks?

Skarlso:
Yeah so I think this pr is blocked until I can inject a secret from a managed secret store via #51.

michelvocks:
> The only two problems are that the token is saved if the pipeline errors out and second, I still have no way of providing and saving a secret which brings us back to #51. :/

What do you mean exactly with "pipeline errors out"? 
Especially when an error happened during the create pipeline step, the github token should not be stored at all. 

> Not going to save the git token as there is, for now, no reason for it. It's only used once, and that's when the pipeline is created and the hook is sent.

I like that a lot until we have #51  implemented! 🤗 

> Modify the UI to have a checkbox that the user want's a webhook, and if it's checked, popup a field where a github TOKEN is requested. ( make apparent that it's not saved at this time ). (( I imagine that this would be a different set of fields for a bitbucket hook. For git a token is enough )).

Awesome! Maybe move the button next to the "Add credentials" button and change the icon (I know this is fine tuning and will be done later 👍 ).

> This means deleting a hook is not handled currently ( that will be a future feature after we save the token in the form of issue #51 (handling secrets via a vault).

That's fine I guess. We also have to think about what happens when the flag "hostname" is changed? We might need to update all webhooks in this case.

> Modify the auto-updater to be opt-in and save a flag for a pipeline saying (autoupdate - true) because maybe you don't always want all your pipeline to update. Thoughts on this?

Like that. We somehow need to integrate that into the create pipeline view. 😄 

> Add a flag called hostname to use for the callback url.

Perfect. ❤️

Skarlso:
>That's fine I guess. We also have to think about what happens when the flag "hostname" is changed? We might need to update all webhooks in this case.

Oh f*ck. :D That didn't occur to me before. :)

rmb938:
Well like jenkins, drone ci, ect... if the admin updates the hostname the hooks just break. No one should try to fix that (because it's hard) so I don't think we should try to.

rmb938:
Also I don't think we need a hostname flag, I know gin has a middleware for getting the request's hostname https://github.com/gin-contrib/location so I'm sure echo has it/won't be hard to add it.

Skarlso:
@rmb938 It does, it just feels hacky to place such a thing on the first whatever request that gets fired off. I don't think it's easily testable, or maintainable. What if the end-point gets moved and the person moving it doesn't realize that there is a critical thing in there for the webhooks. Namely, detect the host and set it. Or even worse, they put it into a different location but that happens to be after a request for webhook creation is fired off.

I think it's just easier and cleaner to have a setting where you explicitly set it, and that's that. No magic, no fuss, no muss. Clean.

That's how I feel about it at least.

rmb938:
Makes sense :)

I don't really have a strong preference one way or another tbh.

michelvocks:
I agree with @Skarlso 😄 
We cannot 100% rely on the detection. And if there is an error or a user has a problem, we can always point to this parameter. 🤗 

But I actually like that we don't support the update stuff for the webhooks. If the hostname changes, all webhooks are broken. Updating can be hard for sure.

Skarlso:
Working finding repository and rebuilding content. :) All fit together nicely.

However it's a bit interesting right now, as the repository I'm testing doesn't really exist yet... I'll find out how to make this better. Maybe really make a pipeline and check out the code.

Skarlso:
So this test failure made me realise something. If I set a service in a test, I can no longer clear it. That's because passing in a `nil` to an interface doesn't actually make it a nil value. It will be an interface with a nil underlying value. But compare it to `nil` and it will be false. That's basic Go stuff, which is much debated, but meh. 

Sooooo, I'm wondering now if the service interface needs a clear service api or if it's fine if I just delete the value I'm setting during the test.

Not sure. Too late... sleepy now.

Skarlso:
Going to investigate some more later. Clearing the store should have worked but it didn't. Nedd to find out why.

Skarlso:
Oh it's because of the hash. Ah.

Skarlso:
Okay so I think we reached the limit of os.Temp. Also the clean service makes no sense. Just use mock (nil). Now with reflect check it works fine.

services/service_provider.go
```diff
@@ -26,7 +28,7 @@ var vaultService security.VaultAPI
 // initialized once in the main.go. If it wouldn't work, main would
 // os.Exit(1) and the rest of the application would just stop.
 func StorageService() (store.GaiaStore, error) {
-	if storeService != nil {
+	if storeService != nil && !reflect.ValueOf(storeService).IsNil() {
```
Skarlso: So the reflect here is because Go's interface is not really just one value. It's a value and a type. So if the mock is set to nil, the before hand nil check won't be saying that it's not nil. It will be of the previous type which won't equal the nil type, thus a simple `!=` won't work. But we still need it because when the system is starting up it is actually nil. :) 

I love go, but this is just insane. :D Oh and it's totally only needed because of the tests. Otherwise we don't really care, because once the application is running, it's running. We don't delete the db from underneath it randomly.

Skarlso:
`testing: open /tmp/go-build997724149/b001/_cover_.out: no such file or directory` huh? :O @michelvocks

Skarlso:
@michelvocks So. Unfortunately because of the service provider, there are now some flakyness to the tests which I'm fixing in this patch. It looks like the services are stepping on each others foot since they are singletons now. :) 

I fixed that by mocking them to nil which in turn made me realise the fact the interfaces are multi-valued behind the back which I totally forgot about. Since most of the time you don't really care.

Skarlso:
Alright, mostly removed the usage of creating `tmp` to `ioutil`. This makes life easier.

services/service_provider.go
michelvocks: Yeah that is one of the bad parts about Go. I think you solved it quite nicely 🤗 GJ

michelvocks:
> So. Unfortunately because of the service provider, there are now some flakyness to the tests which I'm fixing in this patch. It looks like the services are stepping on each others foot since they are singletons now. :)

This hit me also several times. I guess the tests are executed in parallel and this raises sometimes a race condition in the tests? But then I don't understand why the initialisation of the `gaia.Cfg`struct works?

Skarlso:
Naw it's the fact that mostly all providers are using init only if they aren't nil. If the first test happens to delete the db but not reset the service the dB file won't be there because the init doesn't run. In that case defer mock a nil to the service should fix the problem.

Skarlso:
Also the permission denied is because they were still initialized to an earlier state so the next one was looking for the db file in the wrong location.

Skarlso:
@michelvocks I'm realising something here. I can't use the parameters to pipelines for the password. Because I'm only passing in the repository. But that's not the biggest problem. The problem is that in that case you'd have to already have an idea to use webhooks at the moment you create your pipeline because if you have webhooks defined then I need a specific parameter actually to call upon to use the password. I don't know what you decided to use as password value name.

And thus there must be a convention. If you define webhooks you are also required to create a secret for the GITHUB_WEBHOOK_SECRET or something like that.

michelvocks:
> I'm realising something here. I can't use the parameters to pipelines for the password. Because I'm only passing in the repository. 

Uh, you lost me here. 😄 
Why do you actually want to use pipeline parameters to get a password? Is the idea to provide the github token via a pipeline parameter?

> But that's not the biggest problem. The problem is that in that case you'd have to already have an idea to use webhooks at the moment you create your pipeline because if you have webhooks defined then I need a specific parameter actually to call upon to use the password. I don't know what you decided to use as password value name.

Two suggestions here:
1. When you create a pipeline you can just type in a key value. This key value is used to get the real github token from the vault. I suggest to not check if this key/value already exists in the vault during pipeline creation process because the user wants to set it afterwards.
2. In the Settings -> Manage Pipelines view we add the option to add/remove/edit a webhook.

Skarlso:
@michelvocks No it's for the secret which is for payload verification every time an even is received. We aren't saving the token. There is no need for it. It's only ever used once, when the hook is created.

There parameters are useless, because me, in the code, have no idea what key you, as a pipeline builder, picked to be the password, unless I use a convention that says the github webhook payload verification secret MUST be called GITHUB_WEBHOOK_SECRET.

Instead what I AM going to do is if you choose to create a pipeline with a webhook, before the pipeline is created, I'm going to check if in the vault there is a value called GITHUB_WEBHOOK_SECRET. If not, the pipeline isn't allowed to be created because it's not going to work. :) What do you think?

Skarlso:
Or! I'm going to ask for a secret as well. Which I'm going to store in the vault.

So the window where I'm asking for the token will also have a secret to ask for. And the secret is stored in the vault. Yep. This is the best way to do it IMHO. Just ask for a Token + a Secret. :) And store the secret in the vault.

Skarlso:
What do you think ? @michelvocks 
screen shot 2018-08-12 at 16 45 45
screen shot 2018-08-12 at 16 45 41

michelvocks:
What about generating the secret? Does the user really care about this secret?

Skarlso:
@michelvocks I could certainly do that. The only real reason to know the secret is if you wanted to manually test your webhook integration. Without the secret, you wouldn't be able to generate a proper hmac signature. I have no idea how often people would want to manually test their webhooks, but while I was debugging things, I did do it once or twice. 🤷‍♂️

michelvocks:
@Skarlso Generate it and show it once to the user (maybe in the modal panel when you add the webhook?). 🤗

Skarlso:
I don't know how strong Javascript's random generation is. It's probably fine.

Also, I'm thinking that I would want a single password, which then might be used further which also makes a point next to just asking one. Because if I random generate a password for all pipelines, potentially your vault could get very crowded. Like, 20-30-60 webhook passwords on the first pages, might be annoying. :D 

Of course I could always generate it once, and then check if it's already in the vault. But why generate it if the user can just type in something, if it's going to be the task of the frontend anyways. You know what I'm saying. :)

Skarlso:
I choose to generate it in the back once, because:
* You probably will never need it.
* If so, you can actually change the variable in the vault section. :DD

👍

Skarlso:
@michelvocks @rmb938 This is now ready, I think... :)

bidhan-a:
Correct me if I'm wrong but this PR doesn't support non-Github webhooks?

Skarlso:
@bidhan-a You are not wrong. Since non-github webhook require a complete different way of registering them with a different library and api, this PR deals with github specifically. 

Other hooks, like gitlab and bitbucket will be in a separate PR. :)

cmd/gaia/main.go
```diff
@@ -37,6 +37,7 @@ func init() {
 	// command line arguments
 	flag.StringVar(&gaia.Cfg.ListenPort, "port", "8080", "Listen port for gaia")
 	flag.StringVar(&gaia.Cfg.HomePath, "homepath", "", "Path to the gaia home folder")
+	flag.StringVar(&gaia.Cfg.Hostname, "hostname", "https://localhost", "The host's name under which gaia is deployed at exp: https://gaia-pipeline.com")
```
michelvocks: what does exp mean? 😄

frontend/client/views/pipeline/create.vue
```diff
@@ -32,6 +32,12 @@
                   </span>
                   <span>Add credentials</span>
                 </a>
+                <a class="button is-primary" v-on:click="showGitHubWebHookModal">
+                  <span class="icon">
+                    <i class="fa fa-certificate"></i>
```
michelvocks: Would be awesome if you could change the icon for user experience 😃 see https://fontawesome.com/icons?d=gallery most of them are available

michelvocks: Maybe https://fontawesome.com/icons/wrench?style=solid ?+

cmd/gaia/main.go
Skarlso: Example. :D

frontend/client/views/pipeline/create.vue
Skarlso: Sure thing. :) Thanks! :)

cmd/gaia/main.go
michelvocks: maybe e.g.? or for example? 😄

Skarlso: Yeah that's definitely better. :D

Skarlso:
@michelvocks Before doing anything I wanted to run a quick manual. Turns out I had a few bugs in there.. One, I was generating the password at a wrong location. :/ Second, I was sending off the wrong URL for the webhook callback. 🗡 

Unfortunately, because I needed the API version I had to shift the constant into gaia.go where the other constants are lying. That's why there are now more edits.

Skarlso:
@michelvocks umm... dude... The scheduler threw a data race. :/

```bash
2018-08-15T21:47:24.501+0200 [DEBUG] Gaia: error during job execution: error="job failed" job="{2296842551 Job1  [] running}"
panic: send on closed channel

goroutine 13 [running]:
github.com/gaia-pipeline/gaia/scheduler.(*Scheduler).resolveDependencies(0xc420020f00, 0x89e70aca, 0x11e6932, 0x4, 0x0, 0x0, 0xc42000e2d0, 0x1, 0x1, 0x11ea8ba, ...)
        /Users/hannibal/golang/src/github.com/gaia-pipeline/gaia/scheduler/scheduler.go:376 +0x2f0
github.com/gaia-pipeline/gaia/scheduler.(*Scheduler).resolveDependencies(0xc420020f00, 0x8ae70c5d, 0x11e6936, 0x4, 0x0, 0x0, 0xc42000e2d8, 0x1, 0x1, 0x11ea8ba, ...)
        .../golang/src/github.com/gaia-pipeline/gaia/scheduler/scheduler.go:358 +0x223
created by github.com/gaia-pipeline/gaia/scheduler.(*Scheduler).executeScheduler
        ...golang/src/github.com/gaia-pipeline/gaia/scheduler/scheduler.go:429 +0x2c9
```

Skarlso:
To be fair running it with `--race` only failed once out of 10 runs. :/ Which is quiet shitty. But it still failed. So we'll have to take care of that. But to be faire, I'm not going to do this in this pr if you don't mind.

Skarlso:
I'm getting invalid signature on the real deal.. I need to investigate why. :/

Skarlso:
Everything works!! Yiiss!!

Couple of things:
* Password generated wasn't accepted because it had special characters.
* The pipeline was actually not being built because it didn't select the branch which was selected
* The webhook URL was not open to github. It's open now which is safe as it has it's own authentication procedures using `secret` and header checks.

Skarlso:
```bash
2018-08-16T08:39:55.517+0200 [DEBUG] Gaia: checking pipeline: : asdf=<unknown>
2018-08-16T08:39:55.517+0200 [DEBUG] Gaia: selected branch : : refs/heads/dependson=<unknown>
2018-08-16T08:39:56.709+0200 [DEBUG] Gaia: updating pipeline: : asdf=<unknown>
2018-08-16T08:39:58.052+0200 [DEBUG] Gaia: successfully updated: : asdf=<unknown>
```

```bash
POST /api/v1/pipeline/githook  200 OK
```

screen shot 2018-08-16 at 08 42 26

pipeline/git.go
```diff
@@ -148,6 +164,100 @@ func updateAllCurrentPipelines() {
...
+		Name:   github.String("web"),
+		Active: github.Bool(true),
+		Config: config,
+	})
+	if err != nil {
+		gaia.Cfg.Logger.Error("error while trying to create webhook: ", "error", err.Error())
+		return err
+	}
+	gaia.Cfg.Logger.Info("hook created: ", github.Stringify(hook.Name), resp.Status)
+	gaia.Cfg.Logger.Info("hook url: ", hook.GetURL())
```
michelvocks: should be `gaia.Cfg.Logger.Info("hook url ", "url", hook.GetURL())`
Otherwise we get ugly log output like `hook url: : asdf=<unknown>` 😄 
It's always (initial_log_message, key=string, value)

pipeline/git.go
```diff
@@ -85,12 +98,15 @@ func UpdateRepository(pipe *gaia.Pipeline) error {
...
+		ReferenceName: plumbing.ReferenceName(pipe.Repo.SelectedBranch),
+		SingleBranch:  true,
+		RemoteName:    "origin",
+		Auth:          auth,
 	})
 	if err != nil {
 		// It's also an error if the repo is already up to date so we just move on.
-		gaia.Cfg.Logger.Error("error while doing a pull request : ", err.Error())
+		message := "error while doing a pull request: " + err.Error()
+		gaia.Cfg.Logger.Error(message)
```
michelvocks: `gaia.Cfg.Logger.Error("error while doing a pull request", "error", err.Error())`

pipeline/git.go
```diff
@@ -148,6 +164,100 @@ func updateAllCurrentPipelines() {
...
+
+	return GithubClient{
+		Repositories: client.Repositories,
+	}
+}
+
+func createGithubWebhook(token string, repo *gaia.GitRepo, gitRepo GithubRepoService) error {
+	vault, err := services.VaultService(nil)
+	if err != nil {
+		gaia.Cfg.Logger.Error("unable to initialize vault: " + err.Error())
```
michelvocks: `gaia.Cfg.Logger.Error("unable to initialize vault", "error",  err.Error())`

pipeline/git.go
```diff
@@ -148,6 +164,100 @@ func updateAllCurrentPipelines() {
...
+		return err
+	}
+
+	err = vault.LoadSecrets()
+	if err != nil {
+		gaia.Cfg.Logger.Error("unable to open vault: " + err.Error())
```
michelvocks: gaia.Cfg.Logger.Error("unable to open vault", "error", err.Error())

PR #49, opened 2018-07-20
Removed strings.Builder which was a go 1.10 feature. Closes #47

### LINKED ISSUE #47: [Proposal] remove strings.Builder in test cases
The use of `strings.Builder` requires Go 1.10+.  Since this is only used in the test cases, it leaves an interesting situation where I'm able to run Gaia on Go 1.9, but while developing the tests fail.  I think replacing the strings.Builder, which only serves as an `io.Writer` for the logging would normalize the dev/test/prod environments.

Alternatively an update to the README under the Contributing section could clear this up.

Thoughts?

Skarlso:


Skarlso:
No problem. I'm the one causing the trouble after all. :-) Thanks for the merge!

Issue #30, opened 2018-07-12
Modify, Delete Pipeline feature


michelvocks:
We already have a `Settings` subpage. We need a new tab-pane for managing pipelines. 
Maybe a table with all currently active pipelines and every entry has buttons/icons to modify (renaming) and removing this pipeline. 

What needs to be done:

1. Create the functionality in the backend.
2. Create a new tab-pane in the settings view with the required functionality.

bidhan-a:
Hey @michelvocks I would like to take this on if you aren't working on it already. Please let me know!

michelvocks:
Hey @bidhan-a 🤗 
Awesome! Really happy that you want to work on this! ❤️ 
Feel free to start and don't hesitate if you got questions or simply want feedback!

PR #52, opened 2018-07-22
[WIP] Modify, Delete Pipeline feature (#30)

bidhan-a:
Hey @michelvocks 

I have started working on this. 

I've added a new tab pane in the Settings page for managing pipelines. In the backend, I've also added a delete handler which basically does the following: (a) Remove binary, (b) Remove from store, and (c) Remove from the active pipelines list. Is this good enough?

Also, how should we handle the modify/rename thing? Would renaming the binary in the pipelines folder and updating the pipeline in the store & the active pipelines list suffice? Or did you have something else in mind?

Please let me know. 

Thanks!

### DIFF
--- a/handlers/pipeline.go
+++ b/handlers/pipeline.go
@@ -136,6 +136,36 @@ func PipelineGet(c echo.Context) error {
 	return c.String(http.StatusNotFound, errPipelineNotFound.Error())
 }
 
+// PipelineUpdate updates the given pipeline.
+func PipelineUpdate(c echo.Context) error {
+	p := &gaia.Pipeline{}
+	if err := c.Bind(p); err != nil {
+		return c.String(http.StatusBadRequest, err.Error())
+	}
+
+	// Look up pipeline for the given id
+	var foundPipeline *gaia.Pipeline
+	for pipeline := range pipeline.GlobalActivePipelines.Iter() {
+		if pipeline.ID == p.ID {
+			foundPipeline = &pipeline
+		}
+	}
+
+	if foundPipeline == nil {
+		return c.String(http.StatusNotFound, errPipelineNotFound.Error())
+	}
+
+	if foundPipeline.Name != p.Name {
+		// Pipeline name has been changed.
+		err := pipeline.RenameBinary(*foundPipeline, p.Name)
+		if err != nil {
+			return c.String(http.StatusInternalServerError, errPipelineRename.Error())
+		}
+	}
+
+	return c.String(http.StatusOK, "Pipeline has been updated")
+}
+
 // PipelineDelete accepts a pipeline id and deletes it from the
 // store. It also removes the binary inside the pipeline folder.
 func PipelineDelete(c echo.Context) error {
@@ -159,19 +189,19 @@ func PipelineDelete(c echo.Context) error {
 	}
 
 	if foundPipeline == nil {
-		return c.String(http.StatusNotFound, err.Error())
+		return c.String(http.StatusNotFound, errPipelineNotFound.Error())
 	}
 
 	// Delete pipeline binary
 	err = pipeline.DeleteBinary(*foundPipeline)
 	if err != nil {
-		return c.String(http.StatusInternalServerError, err.Error())
+		return c.String(http.StatusInternalServerError, errPipelineDelete.Error())
 	}
 
 	// Delete pipeline from store
 	err = storeService.PipelineDelete(pipelineID)
 	if err != nil {
-		return c.String(http.StatusNotFound, err.Error())
+		return c.String(http.StatusInternalServerError, err.Error())
 	}
 
 	// Remove from active pipelines
--- a/handlers/pipeline.go
+++ b/handlers/pipeline.go
@@ -136,6 +136,50 @@ func PipelineGet(c echo.Context) error {
 	return c.String(http.StatusNotFound, errPipelineNotFound.Error())
 }
 
+// PipelineDelete accepts a pipeline id and deletes it from the
+// store. It also removes the binary inside the pipeline folder.
+func PipelineDelete(c echo.Context) error {
+	pipelineIDStr := c.Param("pipelineid")
+
+	pipelineID, err := strconv.Atoi(pipelineIDStr)
+	if err != nil {
+		return c.String(http.StatusBadRequest, errInvalidPipelineID.Error())
+	}
+
+	// Look up pipeline for the given id
+	var foundPipeline gaia.Pipeline
+	var index int
+	var deletedPipelineIndex int
+	for pipeline := range pipeline.GlobalActivePipelines.Iter() {
+		if pipeline.ID == pipelineID {
+			foundPipeline = pipeline
+			deletedPipelineIndex = index
+		}
+		index++
+	}
+
+	if foundPipeline.Name == "" {
+		return c.String(http.StatusNotFound, err.Error())
+	}
+
+	// Delete pipeline binary
+	err = pipeline.DeleteBinary(foundPipeline)
+	if err != nil {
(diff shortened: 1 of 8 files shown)


handlers/pipeline.go
```diff
@@ -136,6 +136,50 @@ func PipelineGet(c echo.Context) error {
...
+	var deletedPipelineIndex int
+	for pipeline := range pipeline.GlobalActivePipelines.Iter() {
+		if pipeline.ID == pipelineID {
+			foundPipeline = pipeline
+			deletedPipelineIndex = index
+		}
+		index++
+	}
+
+	if foundPipeline.Name == "" {
```
michelvocks: You can make foundPipeline a pointer. Then you can check here: if foundPipeline == nil

michelvocks:
> Also, how should we handle the modify/rename thing? Would renaming the binary in the pipelines folder and updating the pipeline in the store & the active pipelines list suffice? Or did you have something else in mind?

Yes, I think that's enough for now. I was also thinking about changing the source git repository but then you also have to update credentials etc. So for now renaming should be fine! 🤗

handlers/pipeline.go
```diff
@@ -136,6 +136,50 @@ func PipelineGet(c echo.Context) error {
...
+func PipelineDelete(c echo.Context) error {
+	pipelineIDStr := c.Param("pipelineid")
+
+	pipelineID, err := strconv.Atoi(pipelineIDStr)
+	if err != nil {
+		return c.String(http.StatusBadRequest, errInvalidPipelineID.Error())
+	}
+
+	// Look up pipeline for the given id
+	var foundPipeline *gaia.Pipeline
```
Skarlso: There is no need for this to be a reference since you are passing in a copy anyways. Let's keep this on the stack with removing the pointer and the address passing.

handlers/pipeline.go
```diff
@@ -136,6 +136,50 @@ func PipelineGet(c echo.Context) error {
...
+	var index int
+	var deletedPipelineIndex int
+	for pipeline := range pipeline.GlobalActivePipelines.Iter() {
+		if pipeline.ID == pipelineID {
+			foundPipeline = &pipeline
+			deletedPipelineIndex = index
```
Skarlso: Add break here once the pipeline you are looking for is found.

handlers/pipeline.go
```diff
@@ -136,6 +136,50 @@ func PipelineGet(c echo.Context) error {
 	return c.String(http.StatusNotFound, errPipelineNotFound.Error())
 }
 
+// PipelineDelete accepts a pipeline id and deletes it from the
+// store. It also removes the binary inside the pipeline folder.
+func PipelineDelete(c echo.Context) error {
+	pipelineIDStr := c.Param("pipelineid")
```
Skarlso: There is no need for this variable to exist. `c.Param` will return an empty string if the parameter is not found making Atoi fail gracefully. So this can be removed and inlined into Atoi.

handlers/pipeline.go
```diff
@@ -136,6 +136,50 @@ func PipelineGet(c echo.Context) error {
...
+	for pipeline := range pipeline.GlobalActivePipelines.Iter() {
+		if pipeline.ID == pipelineID {
+			foundPipeline = &pipeline
+			deletedPipelineIndex = index
+		}
+		index++
+	}
+
+	if foundPipeline == nil {
+		return c.String(http.StatusNotFound, err.Error())
```
Skarlso: There is a `errPipelineNotFound` error.

handlers/pipeline.go
```diff
@@ -136,6 +136,50 @@ func PipelineGet(c echo.Context) error {
...
+	}
+
+	// Delete pipeline binary
+	err = pipeline.DeleteBinary(*foundPipeline)
+	if err != nil {
+		return c.String(http.StatusInternalServerError, err.Error())
```
Skarlso: Instead of an internal server error I suggest creating an `errPipelineDelete` error.  Something like, `oops sorry, I couldn't delete the binary, maybe permissions are incorrect? Please fix permissions and try again or delete it by hand.`. Thoughts?

handlers/pipeline.go
```diff
@@ -136,6 +136,50 @@ func PipelineGet(c echo.Context) error {
...
+	}
+
+	// Delete pipeline from store
+	err = storeService.PipelineDelete(pipelineID)
+	if err != nil {
+		return c.String(http.StatusNotFound, err.Error())
```
Skarlso: This shouldn't be a `StatusNotFound` not? Rather an internal error since it's a db operation.

handlers/pipeline.go
bidhan-a: Hey @Skarlso I tried adding break inside `Iter()` but after doing that the process just freezes. I assume this is why `break` isn't used here either. https://github.com/gaia-pipeline/gaia/blob/master/pipeline/pipeline.go#L102

handlers/pipeline.go
bidhan-a: I changed it to a reference type as per @michelvocks suggestion.
> You can make foundPipeline a pointer. Then you can check here: if foundPipeline == nil

Thoughts?

handlers/pipeline.go
bidhan-a: I thought so too initially but after seeing this pattern used in other handler functions, I decided to stick with it. :smile:

handlers/pipeline.go
Skarlso: You can check if it is initialized by comparing it to an empty struct of the same type. Rather than a pointer which might be stored on the heap.

Skarlso: I guess it doesn't matter much but avoids a nil pointer expedition.

michelvocks: @bidhan-a Sorry. My mistake here. You implementation was actually right. :sweat_smile:

bidhan-a:
Hey guys @michelvocks @Skarlso 

So I've added the implementation for renaming pipelines. Both renaming & deletion seem to be working properly. Please feel free to critique the code and let me know if I should add/remove anything. If all seems to be okay, I'll start writing test cases for the methods which I've added.

Thanks!

handlers/pipeline.go
```diff
@@ -138,29 +138,45 @@ func PipelineGet(c echo.Context) error {
...
+		p.ExecPath = pipeline.GetExecPath(p)
+
+		// Rename binary
+		err := pipeline.RenameBinary(foundPipeline, p.Name)
 		if err != nil {
 			return c.String(http.StatusInternalServerError, errPipelineRename.Error())
 		}
+
+		// Update pipeline in store
+		err = storeService.PipelinePut(&p)
```
michelvocks: I think it's better to update the name in ``foundPipeline`` and store this object instead of ``p`` which we got from the client. For example, a hacker could potentially change something else than the name in the struct and that will be saved too.

handlers/pipeline.go
```diff
@@ -138,29 +138,45 @@ func PipelineGet(c echo.Context) error {
...
+		if err != nil {
+			return c.String(http.StatusInternalServerError, err.Error())
+		}
+
+		// Update active pipelines
+		pipeline.GlobalActivePipelines.ReplaceByName(foundPipeline.Name, p)
```
michelvocks: Same here. I think we should stick we the old object 😃

michelvocks:
Looks really good now. Good job @bidhan-a 🤗 ❤️ 
Let me know when you are ready with your tests.

bidhan-a:
Thanks @michelvocks . Will let you know once tests are added. :smiley:

bidhan-a:
Hey @michelvocks I've added test cases for all the newly added functions. Let me know if I've missed anything!

PR #54, opened 2018-07-23
Unix sockets to port communication

### LINKED ISSUE #53: Plugin connection -> from unix sockets to port
Currently, Gaia uses unix sockets as communication channel for Gaia <-> Pipeline (Plugin) communication. 

It has the following disadvantages:
* Gaia only runs on Unix (Mac/Linux) systems.
* Adding new SDKs is hard due to new/poor implementation and support of unix sockets.

It has the following advantages:
* It's secure. You can only hijack the communication when you have access to the filesystem. 

We need to change to real port communication. It also means that we have to generate certificates for the communication between Gaia and the Plugin.

michelvocks:
Fixes #53. 
This will be a breaking change. Once this is merged we have to update the SDK immediately otherwise create pipeline will fail. 
For now implemented first version of CA generation and key pair generation which will be signed for mutual Auth.

### DIFF
--- a/scheduler/scheduler_test.go
+++ b/scheduler/scheduler_test.go
@@ -1,6 +1,7 @@
 package scheduler
 
 import (
+	"crypto/tls"
 	"hash/fnv"
 	"os"
 	"os/exec"
@@ -8,6 +9,7 @@ import (
 	"testing"
 
 	"github.com/gaia-pipeline/gaia"
+	"github.com/gaia-pipeline/gaia/security"
 	"github.com/gaia-pipeline/gaia/store"
 	hclog "github.com/hashicorp/go-hclog"
 	uuid "github.com/satori/go.uuid"
@@ -15,11 +17,18 @@ import (
 
 type PluginFake struct{}
 
-func (p *PluginFake) NewPlugin() Plugin                            { return &PluginFake{} }
+func (p *PluginFake) NewPlugin(ca security.CAAPI) Plugin           { return &PluginFake{} }
 func (p *PluginFake) Connect(cmd *exec.Cmd, logPath *string) error { return nil }
 func (p *PluginFake) Execute(j *gaia.Job) error                    { return nil }
 func (p *PluginFake) GetJobs() ([]gaia.Job, error)                 { return prepareJobs(), nil }
 func (p *PluginFake) Close()                                       {}
+
+type CAFake struct{}
+
+func (c *CAFake) CreateSignedCert() (string, string, error)                       { return "", "", nil }
+func (c *CAFake) GenerateTLSConfig(certPath, keyPath string) (*tls.Config, error) { return nil, nil }
+func (c *CAFake) CleanupCerts(crt, key string) error                              { return nil }
+func (c *CAFake) GetCACertPath() (string, string)                                 { return "", "" }
 
 func TestInit(t *testing.T) {
 	gaia.Cfg = &gaia.Config{}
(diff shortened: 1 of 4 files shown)


Skarlso:
Can you put it behind a feature flag of it's breaking everyone? And the flag is off by default and on for anyone opting in? Would it make sense?

michelvocks:
@Skarlso I think I can do that. I just think when we wait longer more users will be pissed at the end. I think now is the perfect time to do changes like these when everyone expects bugs/problems. 😄

Skarlso:
@michelvocks Agreed. :D

michelvocks:
This is now working as intended and make Gaia a bit more secure and flexibe. :tada: :confetti_ball:
I will rebase it after #58 has been merged.

Skarlso:
@michelvocks Yo what's up with this guy? :) The corresponding 58 was merged. :)

michelvocks:
@Skarlso if I merge this, it's not possible to build pipelines anymore when you are running gaia from master branch. It will work again if I merge this: https://github.com/gaia-pipeline/gosdk/pull/1

But then it's not possible to build pipelines in version V0.1.2 which is currently latest released version.

michelvocks:
Ping @Skarlso please have a look and merge if you think so 🤗

Skarlso:
I fixed the merge conflict. 👍 Hope you don't mind. :)

Skarlso:
Interesting. Something new introduced a race condition after merging master. :/

Skarlso:
I tested this on master and it looks okay. So it's this new code plus the new schedule test that's failing.

michelvocks:
@Skarlso I fixed it. Somehow the init method was still in the specific test. No clue what happened there. 😄

Skarlso:
Well done! Awesome.😁

scheduler/scheduler_test.go
```diff
@@ -1,26 +1,35 @@
...
-func (p *PluginFake) NewPlugin() Plugin                            { return &PluginFake{} }
+func (p *PluginFake) NewPlugin(ca security.CAAPI) Plugin           { return &PluginFake{} }
 func (p *PluginFake) Connect(cmd *exec.Cmd, logPath *string) error { return nil }
 func (p *PluginFake) Execute(j *gaia.Job) error                    { return nil }
 func (p *PluginFake) GetJobs() ([]gaia.Job, error)                 { return prepareJobs(), nil }
 func (p *PluginFake) Close()                                       {}
 
+type CAFake struct{}
+
+func (c *CAFake) CreateSignedCert() (string, string, error)                       { return "", "", nil }
```
Skarlso: Okay, so I know that this will be annoying, but... :D Could you please change the API of the CA to named values in the return statement? Because it's annoying that I don't know what `"", "", nil` is actually returning by looking at the pop up of the parameters. :D I don't which is which here. Path, or Cert or KeyPath or whatever, you see what I'm saying?

scheduler/scheduler_test.go
```diff
@@ -36,7 +45,9 @@ func TestInit(t *testing.T) {
 	if err := storeInstance.Init(); err != nil {
 		t.Fatal(err)
 	}
-	s := NewScheduler(storeInstance, &PluginFake{})
+	var ca security.CAAPI
```
Skarlso: you can remove this variable and change the below `ca = &CAFake{}` to simply `ca := &CAFake{}` and all will be well. Do this everywhere. :)

Issue #51, opened 2018-07-21
Secret store for passwords, tokens, api keys and more


michelvocks:
We need a secure way to store passwords, tokens, api keys and everything that should be kept secret.
For example when the user wants to access vault via a pipeline. Usually a token is required to get access to vault and this token should be stored in the secret store. 

After a secret has been stored, it must be possible to automatically inject this password during runtime in the pipeline as a job argument. This is not a problem at all, we just have to decide how the user can setup the relationship between a secret and a pipeline. That means, which secret should be available in which pipeline. 

First option: In the details page create a new view to configure the pipeline. Add a subview to manage secrets.

Second and preferred option: Create a new view below settings. This view is for managing secrets. When a view is created or modified it can be linked to multiple pipelines. This has the advantage that one secret can be shared with multiple pipelines.

What do you think?

michelvocks:
ping @Skarlso

Skarlso:
Picking this up.

michelvocks:
Fixed by #56

Issue #24, opened 2018-07-12
Pipeline parameters


rmb938:
When triggering a pipeline it would be nice to be able to specify input parameters. This would allow users to have dynamic configuration for their pipelines.

An example would be a pipeline that runs database migrations and it needs an input of the database name.

Skarlso:
Hi.

So, a pipeline is just Go code. What prevents you from adding in parameters to your pipeline in whatever means you wish, like a config file or environment property?

Or are you looking for a different solution?

rmb938:
You are correct however there is no way from what I can tell for gaia to pass configuration information when triggering a pipeline.

The configuration I am looking for would only be useful while a single pipeline is ran. So each run of the pipeline can have slightly different inputs.

fubarhouse:
This is _really_ high on my want/need list.

If I'm ever to adopt this as more than an exploration task, it's right up there.

Skarlso:
Oh I didn't know the front-end can't pass in stuff, like environment property values. :)

So, you could, in theory, create a single pipeline and iterate over a config file which contains all the data you need which could fire off a multitude of routines that would show up in the pipeline as things that diverge then come together and create a single view. Fan-out fan-in. 

I realize that's probably not what you are looking for but could very well be a plausible solution for now. Or not? :)

fubarhouse:
@Skarlso Yeah, that is my intention for the time being.

I understand it's no small ask, but it's one of two really important requirements I would need if I'm to ever adopt Gaia as more than a shiny thing in my own time.

The other is writing tasks which can write to the local file system. Unfortunately the only other product which still allows that is Jenkins - so I'm tied down for now.

rmb938:
Yes that could be a temporary solution however having the ability to have input parameters would still be nice. 

@Skarlso I am more than willing to contribute and add this feature if that is a concern.

Skarlso:
Oh for sure, don't get me wrong, I'm not against it. :) Was just trying to come up with a temporary solution so you guys aren't blocked. There is no concern here.

@rmb938 By all means. I'm sure @michelvocks would be delighted. :) In fact, he probably has it on his todo list. :)

michelvocks:
Exactly, this is on my todo list. 😄 
I actually already planned it in the gRPC protocol: https://github.com/gaia-pipeline/protobuf/blob/master/plugin.proto#L15

[...]

PR #56, opened 2018-07-24
A secret vault storage with AES encryption using a certificate

Skarlso:
Deals with #51.

# TODO
- [x] The certificate needs to be obtained and used
- [x] Front-end work
   - [x] Create a view under settings which let's the admin add key/value pairs and then save it
- [x] Write Test for vault handler
- [x] Write documentation for the Vault

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -133,9 +133,9 @@ Gaia will compile it and add it to it's store for later execution.
 Please find a bit more sophisticated example in our `go-example repo`_.
 
 Security
---------
+========
 
-See the Documentation located here: |security-docs|.
+See the Documentation located here: `security-docs`_.
 
 Documentation and more
 ======================
@@ -212,6 +212,7 @@ If you have any questions feel free to contact us on `gitter`_.
 .. _`Kubernetes deployment with vault integration`: https://docs.gaia-pipeline.io/tutorials/kube-vault-deploy/
 .. _`git push`: https://git-scm.com/docs/git-push
 .. _`HTTP/2`: https://http2.github.io/
+.. _`security-docs`: https://github.com/gaia-pipeline/gaia/blob/master/security/README.md
 
 .. |build-status| image:: https://circleci.com/gh/gaia-pipeline/gaia/tree/master.svg?style=shield&circle-token=c0e15edfb08f8076076cbbb55558af6cfecb89b8
     :alt: Build Status
@@ -264,6 +265,4 @@ If you have any questions feel free to contact us on `gitter`_.
 .. |sh-settings| image:: https://cdn.rawgit.com/michelvocks/6868118d0da06a422e69e453497eb30d/raw/142a2969c4d27d4135ef8f96213bb166009fda1e/settings.png
     :alt: gaia settings screenshot
     :width: 650px
-.. |security-docs|
-    :alt: Security in Gaia
-    :target: https://github.com/gaia-pipeline/gaia/blob/master/security/README.md
+
--- a/README.rst
+++ b/README.rst
@@ -1,5 +1,5 @@
 .. raw:: html
-    
+
     <img src="https://cdn.rawgit.com/michelvocks/ef3894f63c3bb004bca1a2fd5f7eb644/raw/c36d614db8afe229b466b38de1636a82ad809f64/gaia-logo-text.png" width="650px">
 
 |build-status| |go-report| |go-doc| |apache2| |chat| |codecov|
@@ -13,8 +13,8 @@ Motivation
 
 .. begin-motivation
 
-*Automation Engineer*, *DevOps*, *SRE*, *Cloud Engineer*, 
-*Platform Engineer* - they all have one in common: 
+*Automation Engineer*, *DevOps*, *SRE*, *Cloud Engineer*,
+*Platform Engineer* - they all have one in common:
 The majority of tech people are not motivated to take up this work and they are hard to recruit.
 
 One of the main reasons for this is the abstraction and poor execution of many automation tools. They come with their own configuration (`YAML`_ syntax) specification or limit the user to one specific programming language. Testing is nearly impossible because most automation tools lack the ability to mock services and subsystems. Even tiny things, for example parsing a JSON file, are sometimes really painful because external, outdated libraries were used and not included in the standard framework.
@@ -26,11 +26,11 @@ How does it work?
 
 .. begin-architecture
 
-Gaia is based on `HashiCorp's go-plugin`_. It's a plugin system that uses `gRPC`_ to communicate over `HTTP/2`_. HashiCorp developed this tool initially for `Packer`_ but it's now heavily used by `Terraform`_, `Nomad`_, and `Vault`_ too. 
+Gaia is based on `HashiCorp's go-plugin`_. It's a plugin system that uses `gRPC`_ to communicate over `HTTP/2`_. HashiCorp developed this tool initially for `Packer`_ but it's now heavily used by `Terraform`_, `Nomad`_, and `Vault`_ too.
 
 Plugins, which we named pipelines, are applications which can be written in any programming language as long as `gRPC`_ is supported. All functions, which we call Jobs, are exposed to Gaia and can form up a dependency graph which describes the order of execution.
 
-Pipelines can be compiled locally or simply over the build system. Gaia clones the git repository and automatically builds the included pipeline. If a change (`git push`_) happened, Gaia will automatically rebuild the pipeline for you.  
+Pipelines can be compiled locally or simply over the build system. Gaia clones the git repository and automatically builds the included pipeline. If a change (`git push`_) happened, Gaia will automatically rebuild the pipeline for you.
 
 After a pipeline has been started, all log output are returned back to Gaia and displayed in a detailed overview with their final result status.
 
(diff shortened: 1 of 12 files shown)


security/vault_test.go
```diff
@@ -0,0 +1,191 @@
...
+		Output: buf,
+		Name:   "Gaia",
+	})
+	v, err := NewVault()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.Remove(vaultName)
+	v.Cert = []byte("test")
+	v.OpenVault()
```
Skarlso: This isn't needed here.

Skarlso:
@michelvocks The view of the vault. :)

screen shot 2018-07-24 at 22 21 29

handlers/vault.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		gaia.Cfg.Logger.Error("error reading secret", "error", err.Error())
+		return c.String(http.StatusBadRequest, err.Error())
+	}
+	v, err := security.NewVault()
+	if err != nil {
+		gaia.Cfg.Logger.Error("error initializing vault", "error", err.Error())
+		return c.String(http.StatusInternalServerError, err.Error())
+	}
+	v.Add(s.Key, []byte(s.Value))
```
Skarlso: I'm missing an open here so this is overwriting whatever is in the vault at the moment. Which brings me to rename the functions to make it more apparent how this needs to be used.

Skarlso:
I need to write docs for the vault.

Skarlso:
Working secret list.

screen shot 2018-07-25 at 07 57 50

michelvocks:
Is the value as plain text? :anguished:
I think we should hide that.

Skarlso:
@michelvocks for sure. It's for testing purposes only. :)

Skarlso:
@michelvocks masked :)

screen shot 2018-07-25 at 21 43 06

Skarlso:
Dooooooneee!

Vault empty:
screen shot 2018-07-25 at 23 35 23


Vault with secret:
screen shot 2018-07-25 at 23 35 42
screen shot 2018-07-25 at 23 35 50

Vault editing secret:
screen shot 2018-07-25 at 23 36 10

Vault deleting secret:
screen shot 2018-07-25 at 23 36 19

Skarlso:
Ah, I need to write docs for this.

handlers/vault.go
```diff
@@ -0,0 +1,122 @@
...
+	err = v.SaveSecrets()
+	if err != nil {
+		gaia.Cfg.Logger.Error("error saving vault", "error", err.Error())
+		return c.String(http.StatusInternalServerError, err.Error())
+	}
+	gaia.Cfg.Logger.Info("secret successfully added")
+	return c.String(http.StatusOK, "secret successfully added")
```
michelvocks: To be honest, I'm not really good at HTTP Status codes but this should be Created, right? 😆

security/vault.go
```diff
@@ -0,0 +1,242 @@
...
+		gaia.Cfg.Logger.Info("vault file doesn't exist. creating...")
+		_, err := os.Create(vaultPath)
+		if err != nil {
+			gaia.Cfg.Logger.Error("failed creating vault file: ", err.Error())
+			return nil, err
+		}
+	}
+
+	// Setting up certificate key content
+	c, err := InitCA()
```
michelvocks: Don't create the CA here. It will be already created in the main.go file. You should pass this instance via Parameter to `NewVault`.

security/vault.go
```diff
@@ -0,0 +1,242 @@
...
+	if err != nil {
+		return nil, err
+	}
+	data, err := ioutil.ReadFile(c.caKeyPath)
```
michelvocks: You should not access the private variable `caKeyPath` here. Use the method `GetCACertPath`.

security/vault.go
```diff
@@ -0,0 +1,242 @@
...
+		return err
+	}
+	err = ioutil.WriteFile(v.Path, []byte(encryptedData), 0400)
+	return err
+}
+
+// GetAll returns all keys and values in a copy of the internal data.
+func (v *Vault) GetAll() []string {
+	m := make([]string, 0)
+	for k := range v.data {
```
michelvocks: What happens when someone accesses `Add` and in the same time `GetAll`? Isn't this a race condition?

security/vault.go
```diff
@@ -0,0 +1,242 @@
...
+	"github.com/gaia-pipeline/gaia"
+)
+
+const (
+	vaultName = ".gaia_vault"
+)
+
+// Vault is a secret storage for data that gaia needs to store encrypted.
+type Vault struct {
+	Path string
```
michelvocks: Is it on purpose that Path and Cert are global available?

frontend/client/views/vault/index.vue
```diff
@@ -0,0 +1,410 @@
...
+
+      this.$http
+        .post('/api/v1/secret/update', this.selectSecret)
+        .then(response => {
+          openNotification({
+            title: 'Secret changed!',
+            message: 'Secret has been successful changed.',
+            type: 'success'
+          })
+          this.selectSecret.newvalue = ''
```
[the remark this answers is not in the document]
Skarlso: didn't know that, thanks! :)

handlers/vault.go
Skarlso: Ops, absolutely.

security/vault.go
Skarlso: Yeah, but I'm always using NewVault, and Ca won't be something global, and it shouldn't be. A call to init, should just be idempotent, OR... be a singleton which still is a global 🤔 Not sure.

security/vault.go
Skarlso: Ops, I mean to do that, huh.

security/vault.go
Skarlso: Yes, there should be a Read Lock here. :/ Thanks.

security/vault.go
Skarlso: You mean exported? Hum, haven't really thought about it, but you are right. They aren't used directly, so they can be private for only internal use. Thanks. :)

handlers/vault.go
```diff
@@ -0,0 +1,122 @@
...
+	Value string `json:"value"`
+}
+
+type updateSecret struct {
+	Key   string `json:"key"`
+	Value string `json:"newvalue"`
+}
+
+// UpdateSecret updates a secret using the vault.
+func UpdateSecret(c echo.Context) error {
```
Skarlso: I don't like this duplication. I'm going to pull it into SetSecret which updates or Creates.

frontend/client/views/vault/index.vue
```diff
@@ -0,0 +1,410 @@
...
+    close () {
+      this.showEditSecretModal = false
+      this.showDeleteSecretModal = false
+      this.showAddSecretModal = false
+      this.selectSecret = {}
+      this.$emit('close')
+    },
+
+    changeSecret () {
+      // pre-validate
```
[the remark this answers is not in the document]
Skarlso: @michelvocks I suck at Javascript. How do I do that? :D

README.rst
```diff
@@ -132,6 +132,11 @@ Gaia will compile it and add it to it's store for later execution.
 
 Please find a bit more sophisticated example in our `go-example repo`_.
 
+Security
+--------
+
+See the Documentation located here: |security-docs|.
```
Skarlso: this isn't linking the way I thought it would :D

Skarlso:
@michelvocks Battle tested the front-end with the Javascript modifications and all worked fine. Thanks for that. :)

Also updated the docs, everything should be cool now. Thanks for those fixes.

Skarlso:
@michelvocks A little change... I slept on this and realised that we might want to change the vault's storage medium at some point to a database, or I don't know what.

In that case, I created the VaultStorer interface here: https://github.com/gaia-pipeline/gaia/pull/56/files#diff-6d277307688fa8de2b326d79a993e6b4R25

This made me able to mock the vault to a simple `[]byte` storage medium in the tests, which is awesome. I'm going to do this in #61 for others as well.

michelvocks:
Awesome! 🤗 
I think we shouldn't merge this PR as long as #24 isn't implemented. Currently, you can manage secrets but they are not useful. What do you think, @Skarlso?

Skarlso:
I think that PR would be better off if this one is already in, since we could immediately utilise the functionality in #48. Which is blocked on the both of them. Also, I would rather start on the parameter stuff once this and #61 are in because they provide facilities and implementation details that could prove useful. Especially the Service Provider. In fact, I kind of want that in first, since it touches everything basically.

And if this is in, I can also make the vault part of the service provider.

At least that's what I'm thinking @michelvocks. :)

michelvocks:
Good point. Could you please squash your commits? Makes the history a bit cleaner. 🤗 

Can you also have a look at the following change? https://github.com/gaia-pipeline/gaia/pull/60/files#diff-1e72a85e5dbe49ee4428eae35d6f1ee6R207
Instead of using the PKCS#1 format we need to use the PKCS#8 format. Does it affect this PR?

Skarlso:
It does, yeah. But not that much and I agree with you. I'll do the changes. Also git can squash commits. You can select the squash and merge when merging, but I can do that if you wish ofc. :-)

Skarlso:
So, I completely misunderstood this comment, and thought about something completely different. :D yay, me. 

So as I wrote in Slack, but I'm writing it here too, what's happening with the certificate is, that I'm deriving a 32 bit password from it by base64-ing the content, and chopping it into a aes.BlockSize'd slice.

This means that the vault doesn't care what is in there. I'm also applying some padding in case of testing. The IV is generated and then used upon decryption too. It's marshalled into the encrypted data.

All in all it should be pretty secure. And it's AES 256 so it should be safe from brute forcing.

PR #58, opened 2018-07-25
Refactored security package

michelvocks:
Refactored security package. Currently, the CA will be initiated and not used. 

Ping @Skarlso

### DIFF
--- a/security/ca.go
+++ b/security/ca.go
@@ -7,6 +7,7 @@ import (
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/pem"
+	"errors"
 	"io/ioutil"
 	"math/big"
 	"os"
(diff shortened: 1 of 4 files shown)


security/ca.go
```diff
@@ -26,16 +27,66 @@ const (
...
+	certExist, keyExist := false, false
+	if _, err := os.Stat(c.caCertPath); err == nil {
+		certExist = true
+	}
+	if _, err := os.Stat(c.caKeyPath); err == nil {
+		keyExist = true
+	}
+
+	// Both exist, skip
+	if certExist && keyExist {
```
Skarlso: This shloud be simplified to

```go
_, certErr := os.Stat(c.caCertPath)
_, certPathErr := os.Stat(c.caKeyPath)

if certErr == nil && certPathErr == nil {
return
}
```

security/ca.go
```diff
@@ -155,10 +203,44 @@ func createSignedCert() (string, string, error) {
...
+
+// CleanupCerts removes certificates at the given path.
+func (c *CA) CleanupCerts(crt, key string) error {
 	if err := os.Remove(crt); err != nil {
 		return err
 	}
 	return os.Remove(key)
 }
+
+// GetCACertPath returns the path to the cert from the root CA.
```
Skarlso: Isn't the comment wrong here? It also returns the key's path.

michelvocks:
Good points @Skarlso 🤗

Issue #44, opened 2018-07-19
Refactor Git module to use a Storer


Skarlso:
Right now, the git module can only be tested via actually cloning stuff.

I would like to avoid that, but my current PR is convoluted enough. :) I'm going to do that refactoring with in the confines of this ticket.

michelvocks:
What's bad about cloning stuff? 😄 
I mean, that's the real use-case and we can easily test it. What are the disadvantages in your opinion? 🤗

Skarlso:
Generally it's frowned upon if you poll github a lot and Gods know they need a break. :-) we need end to end tests don't get me wrong. One or more but this scenario needs a push. Which means everytime someone runs this it will pull and push into a repo which is bad. We could mske this test with a switch to only run in Ci but better would be if all it would do is do it in memory.

Skarlso:
This is just like the store needs to be an interface in order to test stuff at the unit level instead of using a really bolt db. And have some end to end tests that actually use bolt db. :-)

michelvocks:
Well. I know from other big projects they are also pulling something (docker images, files from servers etc.) and a git clone is basically the same. We are downloading files from a server. I don't know if github has some kind of DDoS protection which will, at some point, lock you out but I'm pretty sure you will never hit that locally. 😄 

Why do we need a push test? We do not provide any push functionality.

Skarlso:
Ah so that I can test the pull on repo change functionality of the update pipeline on code change feature. :-)

Well, just because other people do it doesn't mean Gaia has to do it. :-)

Generally I think that a unit test is a test where your have full control over all the moving parts. An integration test is where you have a third party of any kind. Generally what I don't want is a test failing because github happens to not be available or throttling us. An integration test is fine frol failing that way. But not a unit test. A unit test should run without Internet in the extremes.

michelvocks:
Alright, good point. Didn't thought about the pull on repo change test. 😄

Skarlso:
Fair enough, it's a new feature after all. :) 👍 Thanks. :)

Skarlso:
This might not be an issue any longer as I found a way to nicely get around the github interface. I can do the same with git as well.

Skarlso:
In test:

https://github.com/gaia-pipeline/gaia/blob/master/pipeline/git_test.go#L252

Using this pattern to mock out a certain part of an API by wrapping it into an interface:

https://github.com/gaia-pipeline/gaia/blob/master/pipeline/git.go#L168-L194

PR #62, opened 2018-07-28
Use auth info (if present) when updating pipelines

bidhan-a:
I was getting a lot of these errors since my pipelines are in a git repo which requires authetication

screenshot

This PR should fix that.

Please let me know your thoughts.

Thanks!

### DIFF
--- a/pipeline/git_test.go
+++ b/pipeline/git_test.go
@@ -156,3 +156,60 @@ func TestUpdateAllPipelinesHundredPipelines(t *testing.T) {
 		t.Fatal("log output did not contain error message that the repo is up-to-date.: ", b.String())
 	}
 }
+
+func TestGetAuthInfo(t *testing.T) {
+	repoWithUsernameAndPassword := &gaia.GitRepo{
+		URL:       "https://github.com/gaia-pipeline/go-test-example",
+		LocalDest: "tmp",
+		Username:  "username",
+		Password:  "password",
+	}
+
+	auth, err := getAuthInfo(repoWithUsernameAndPassword)
+	if auth == nil {
+		t.Fatal("auth should not be nil when username and password is provided")
+	}
+
+	samplePrivateKey := `
+-----BEGIN RSA PRIVATE KEY-----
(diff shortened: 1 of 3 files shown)


Skarlso:
Looks good! Can you add tests please? :) The policy is, that no code change or code goes in without tests. :)

Skarlso:
@bidhan-a hmm, you'll have to take a look at the git repo framework. see how they do it in their test cases.

Ultimately you can do an integration test with actual credentials. Create a test user and test repo for it. But that would a last resort I think.

Otherwise, this might be a good chance to do #44 :D

bidhan-a:
I couldn't find any test case in go-git which handles repos that require authentication :confused:

It seems like we'll have to look into other options.

Skarlso:
@bidhan-a Oh wait. I think we actually don't have to. Just test this function in the three ways:
https://github.com/gaia-pipeline/gaia/pull/62/files#diff-8b0b0b31212c3aeba36c22017d34614dR144

First, if you provide a username and password the authentication will be that. If not, than there is a public key auth and if neither then nil, nil is returned. We don't have to test github authentication process in this PR because frankly, we don't care. It's a third party component.

We can think about an integration test, but that's minuscule compared to three things we can test now. How does that sound?

bidhan-a:
@Skarlso Yep, that sounds good to me. Thanks! :grin:

pipeline/git_test.go
```diff
@@ -156,3 +156,60 @@ func TestUpdateAllPipelinesHundredPipelines(t *testing.T) {
 		t.Fatal("log output did not contain error message that the repo is up-to-date.: ", b.String())
 	}
 }
+
+func TestGetAuthInfo(t *testing.T) {
```
Skarlso: Would you be so kind as to extract these into three smaller tests? Or, if you want the setup, than use `t.Run("bla", func() {})` in order to have three legs in one test while still using the setups.

michelvocks:
Good job, @bidhan-a  🤗 In my opinion this is ready to merge. @bidhan-a @Skarlso any objections?

PR #66, opened 2018-07-30
Fixed flaky schedule test

### LINKED ISSUE #45: Refactor the scheduler test because it's flaky
Smells:

scheduler_test.go:144:146
```go
	// Wait some time to pickup work and finish.
	// We have to wait at least 3 seconds for scheduler tick interval.
	time.Sleep(5 * time.Second)
```

This just failed with the error that the job is in an invalid status. I think this should be fortified or it will cause problems in the future. The scheduler needs to be tested in a different way, or manually kicked off at this stage, or somehow signal on a channel that it's done with its round.

michelvocks:
Fixes #45 . Ping @Skarlso

### DIFF
--- a/scheduler/scheduler_test.go
+++ b/scheduler/scheduler_test.go
@@ -6,7 +6,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"testing"
-	"time"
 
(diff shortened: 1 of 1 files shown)


Skarlso:
LGTM :)

PR #70, opened 2018-08-05
Refactoring and test fixes.

Skarlso:
This PR fixes the following: 

* When tests are sharing services they must clean up after themselves or the next test will use an already initialised service. This means that things like, the location of the database might be cached, or if the service is mocked out, it might not work with the next test. Therefor a `defer func() { services.MockStorageService(nil)}()` is needed for cleanup.
* Some tests are deleting the database file. The services though are only initialised if the underlying service singleton is `nil`. If not, it will just return that, but will fail never the less because the file doesn't exists like the db, or the certificate. Again, a set to nil will fix this problem.
* `os.TempDir` isn't unique enough. There are cases now that the tests step on each other. `ioutil.TempDir("", "TestName")` must be used to ensure semi uniqueness. Maybe add a random number in there.
* Minimal change to use an even smaller repository in tests which only contains fmt instead of the whole world as dependency which made the test slow and clunky.

There might be more cases when running a test something doesn't work. A good first step is add in a defer mock to nil for services that the code might use. Like, vault, or storage.

Also, pro tip... once in a while run `go clean -testcache`! This will erase every cached content and force tests to run. Might surprise you that some tests start to fail even though you though they are passing. Watch out for this tid-bit: `ok      github.com/gaia-pipeline/gaia/handlers  **(cached)**`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ GO_LDFLAGS_STATIC=-ldflags "-s -w -extldflags -static"
 default: dev
 
 dev:
-	go run ./cmd/gaia/main.go -homepath=${PWD}/tmp -dev=true -poll=true
+	go run ./cmd/gaia/main.go -homepath=${PWD}/tmp -dev=true
 
 compile_frontend:
 	cd ./frontend && \
--- a/handlers/pipeline_test.go
+++ b/handlers/pipeline_test.go
@@ -32,14 +32,10 @@ func (ms *mockScheduleService) SchedulePipeline(p *gaia.Pipeline) (*gaia.Pipelin
 }
 
 func TestPipelineGitLSRemote(t *testing.T) {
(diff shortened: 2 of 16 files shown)


Skarlso:
Ready to Merge, of course only after review. :)

Makefile
```diff
@@ -4,7 +4,7 @@ GO_LDFLAGS_STATIC=-ldflags "-s -w -extldflags -static"
 default: dev
 
 dev:
-	go run ./cmd/gaia/main.go -homepath=${PWD}/tmp -dev=true -poll=true
+	go run ./cmd/gaia/main.go -homepath=${PWD}/tmp -dev=true
```
michelvocks: I guess we already talked about this. I think we can leave it in. What do you think? 🤗

michelvocks: I will just merge it in now. We can add it later again if decision has been made 🤗

PR #71, opened 2018-08-09
Pipeline parameters

### LINKED ISSUE #24: Pipeline parameters

michelvocks:
Fixes #24

### DIFF
--- a/cmd/gaia/main.go
+++ b/cmd/gaia/main.go
@@ -146,6 +146,18 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Initiating Vault
+	// Check Vault path
+	if gaia.Cfg.VaultPath == "" {
+		// Set default to data folder
+		gaia.Cfg.VaultPath = gaia.Cfg.DataPath
+	}
+	_, err = services.VaultService(nil)
+	if err != nil {
+		gaia.Cfg.Logger.Error("error initiating vault")
+		os.Exit(1)
+	}
+
 	// Initialize scheduler
 	_, err = services.SchedulerService()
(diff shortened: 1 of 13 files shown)


michelvocks:
ping @Skarlso 
It's ready for review. We should merge #67 before so this PR doesn't include it anymore.
For testing you can use following pipeline: https://github.com/michelvocks/gaia-test-pipeline/tree/dependson

Skarlso:
Excellent! Looks good on a first run, I'll give it a manual. :)

Skarlso:
So, unfortunately I ran into a couple of problems @michelvocks. :/

I created two pipelines with the same repository. Once I set values to one, the other just flashed in the argument window and then puff, made it disappear immediately and went on running the pipeline. It came out as empty value in the logs.

Second, the variables aren't cleared after once specifying them. Meaning, once you defined them, you can't every define them again, because they are already saved in cache I think, and so the asking module just flashes in and then immediately disappears.

The Vault stuff is working nicely however. :)

michelvocks:
Oh thanks for the report. Let me try to reproduce and fix it 🤗

michelvocks:
@Skarlso I rebased and fixed your reported bugs. Try it again and let me know 🤗 
I also updated my test pipeline to point back to gaia-pipeline repositories: https://github.com/michelvocks/gaia-test-pipeline/tree/master

PR #78, opened 2018-08-17
Implemented python support

### LINKED ISSUE #68: Python language Support
Gaia Pipelines in python support.

michelvocks:
Fixes #68 

This feature will set the minimum golang version to 1.10 because I had to switch to `x509.MarshalPKCS8PrivateKey` which was introduced with go 1.10.

### DIFF
--- a/pipeline/update_pipeline.go
+++ b/pipeline/update_pipeline.go
@@ -1,0 +1,46 @@
+package pipeline
+
+import (
+	"errors"
+	"os"
+	"os/exec"
+	"path/filepath"
+
+	"github.com/gaia-pipeline/gaia"
+)
+
+// updatePipeline executes update steps dependent on the pipeline type.
+// Some pipeline types may don't require this.
+func updatePipeline(p *gaia.Pipeline) error {
+	switch p.Type {
+	case gaia.PTypePython:
+		// Remove virtual environment if existend
+		virtualEnvPath := filepath.Join(gaia.Cfg.HomePath, gaia.TmpFolder, gaia.TmpPythonFolder, p.Name)
+		os.RemoveAll(virtualEnvPath)
+
+		// Create virtual environment
+		path, err := exec.LookPath("virtualenv")
+		if err != nil {
+			return errors.New("cannot find virtualenv executeable")
+		}
+		cmd := exec.Command(path, virtualEnvPath)
+		if err := cmd.Run(); err != nil {
+			return err
+		}
+
+		// copy distribution file to environment and remove pipeline type at the end.
+		// we have to do this otherwise pip will fail.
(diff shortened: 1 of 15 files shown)


michelvocks:
@Skarlso This should be fine like it is.

This was a bit more complicated compared to go and java and I had to add some extra stuff like extracting the pipelines in their own virtual environment. 
If you want to test it you can use my python test pipeline: https://github.com/michelvocks/gaia-test-pipeline-python
 *Important* You need python 2.7+, pip (`easy_install pip`), and virtualenv (pip install virtualenv).

Also have a look at the SDK if you want: https://github.com/gaia-pipeline/pythonsdk

pipeline/update_pipeline.go
```diff
@@ -0,0 +1,56 @@
...
+
+		// copy distribution file to environment and remove pipeline type at the end.
+		// we have to do this otherwise pip will fail.
+		err = copyFileContents(p.ExecPath, filepath.Join(virtualEnvPath, p.Name+".tar.gz"))
+		if err != nil {
+			return err
+		}
+
+		// install plugin in this environment
+		cmd = exec.Command("/bin/sh", "-c", fmt.Sprintf(pythonPipInstallCmd, filepath.Join(virtualEnvPath, p.Name)))
```
Skarlso: Uh. This will make supporting windows harder.

scheduler/scheduler.go
```diff
@@ -625,6 +629,15 @@ func createPipelineCmd(p *gaia.Pipeline) *exec.Cmd {
 			"-jar",
 			p.ExecPath,
 		}
+	case gaia.PTypePython:
+		// Build start command
+		c.Path = "/bin/sh"
+		c.Args = []string{
+			"/bin/sh",
```
Skarlso: This too.

Not a blocker for now, but we might want to start thinking about abstracting this.

Even if you put it into a function to compartmentalise it, it will be easier to change later? Cause it's on place only.

security/ca.go
```diff
@@ -204,7 +203,7 @@ func (c *CA) CreateSignedCert() (string, string, error) {
 	}
 
 	// Write out key file in PKCS#8 format
-	privateKey, err := marshalPKCS8PrivateKey(priv)
+	privateKey, err := x509.MarshalPKCS8PrivateKey(priv)
```
Skarlso: Noice! :)

pipeline/update_pipeline.go
michelvocks: True. I guess we have to add later a decision making here: e.g. if windows do this otherwise use this.

scheduler/scheduler.go
michelvocks: I think we can refactor this later. I don't want to hold back the new release because of this 🤗

Issue #32, opened 2018-07-13
Abiltiy to set configuration flags via environment variables


rmb938:
When deploying gaia it would be nice to be able to set the configuration flags via environment variables. 

For ease of development it may be useful to use a cli/configuration system like https://github.com/spf13/viper or https://github.com/alecthomas/kingpin

michelvocks:
We also need to have a look at the configuration in general. I think we will get more configuration parameters in the future and passing in parameters will be unmanageable. :skull:

Not quite sure but do you know what other big open source projects use?

rmb938:
I typically see viper being used in other projects, it can do cli flags, env vars and config files. 

I don't know ifhaving tons of cli flags is a bad thing though. Just look at docker and kubernetes, mostly everyone is fine with having tons of flags.

michelvocks:
This looks actually really good. It's simple and self-explained. I think we can just use it and when we need to do modifications we fork it. What do you think @Skarlso ?

Skarlso:
@rmb938 Implemented. Hopefully that fits the desired need, without also having a major impact code wise.

rmb938:
Yup looks good. Thanks :D

PR #93, opened 2018-08-26
Configuration is updated to use files or environment variables.

Skarlso:


README.rst
```diff
@@ -83,6 +83,45 @@ Gaia will automatically detect the folder of the binary and will place all data
...
+
+2. Environment Properties
+
+Environment variables can be used to overwrite any configuration settings / defaults that are already defined in Gaia.
+
+All Environment variables must be pre-fixed with `GAIA_` in order to avoid collision with other variables.
+
+For example:
+
+.. code:: bash
```
Skarlso: @michelvocks How do I format code? :O This isn't showing up. :O

Skarlso:
Deals with #32

README.rst
michelvocks: You need to add an empty line after `.. code:: bash` for example:

.. code: bash

    #!/bin/bash
    echo hello world

Skarlso: Ah, thanks 🙇 :)

michelvocks:
What do you think about moving the documentation part out of the README.rst file and into the [docs](https://docs.gaia-pipeline.io/)? 😉 
This is information is important but I'm not sure if it is really necessary to add this to the README.rst file? Otherwise LGTM 🤗

Skarlso:
@michelvocks Hum. Where do you think I should put it? Should I create a new section? 
Should I add it here? https://github.com/gaia-pipeline/docs/blob/master/content/getting-started/install-gaia.md

michelvocks:
https://github.com/gaia-pipeline/docs/blob/master/content/getting-started/install-gaia.md is fine I guess. 😄

PR #96, opened 2018-08-29
Implemented pipeline integrity check during pipeline creation process

### LINKED ISSUE #91: Inform user when pipeline is broken
When a new pipeline has been created and detected by Gaia, it usually tries to start the pipeline and get all jobs for the detailed overview: https://github.com/gaia-pipeline/gaia/blob/master/pipeline/ticker.go#L114

Sometimes this can fail when:
* The pipeline throws runtime exceptions (e.g. dependencies are not correctly declared).
* The binary cannot be started because a dependency is missing.
* Memory limit exceeded.
* Many more reasons what can actually go wrong... 😄 

We should make this more resilient and inform the user when something goes wrong.

michelvocks:
Fixes #91 

Implemented an integrity check during pipeline creation process. Needs to be tested.

### DIFF
--- a/frontend/client/views/overview/index.vue
+++ b/frontend/client/views/overview/index.vue
@@ -1,7 +1,7 @@
 <template>
   <div class="columns is-multiline">
     <template v-for="(pipeline, index) in pipelines">
-      <div class="column is-one-third" :key="index">
+      <div class="column is-one-third" :key="index" v-if="!pipeline.p.notvalid">
         <div class="pipeline-box notification content-article">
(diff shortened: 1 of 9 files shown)


michelvocks:
@Skarlso This improves a bit the user experience when the pipeline is broken 😄 
You can try it with this broken pipeline: https://github.com/michelvocks/gaia-test-pipeline
I just added a dependency which does not exist.

Skarlso:
LGTM! 👍

PR #98, opened 2018-08-30
Add, initial addition of crystal assets

bneil:
Basically a copied version of the python pipelines with some slight
modifications and logo additions from the crystal lang site.


michelvocks:
Thanks for your contribution, @bneil 🤗 
It seems like crystel-lang does not support HTTP/2 nor gRPC yet: https://github.com/crystal-lang/crystal/wiki/Roadmap#standard-library
This is actually a requirement for Gaia to support pipelines in that specific language. 
Otherwise the go-plugin cannot communicate with the pipeline 😞

PR #101, opened 2018-09-03
Added tests for plugin package. Implemented job results feature.

### LINKED ISSUE #100: Fail pipeline when job returns an error
Gaia should stop the pipeline and mark it as failed when a job returns or throws an error. This is currently not implemented.

michelvocks:
Fixes #100 and also #2


michelvocks:
Ping @Skarlso

Skarlso:
How is this a coverage decrease? :D

Skarlso:
Oh! Job results feature. :P

Skarlso:
\o/ for client tests.

michelvocks:
@Skarlso Yeah it seems like when you have no tests at all then it doesn't count at all. :man_shrugging:

Skarlso:
Hhhaha. Going to pick this up for a manual later today and merge it when everything is okay. :)

scheduler/scheduler.go
```diff
@@ -322,14 +332,8 @@ func executeJob(j gaia.Job, pS Plugin, triggerSave chan gaia.Job) {
 
 	// Execute job
 	if err := pS.Execute(&j); err != nil {
-		// TODO: Show it to user
 		gaia.Cfg.Logger.Debug("error during job execution", "error", err.Error(), "job", j)
-		j.Status = gaia.JobFailed
```
Skarlso: I'm pretty sure you want to set the job status to failed here? How come you deleted all this code? If you don't have to return any more.

scheduler/scheduler.go
```diff
@@ -48,9 +48,12 @@ type Plugin interface {
 	// NewPlugin creates a new instance of plugin
 	NewPlugin(ca security.CAAPI) Plugin
 
-	// Connect initializes the connection with the execution command
-	// and the log path wbere the logs should be stored.
-	Connect(command *exec.Cmd, logPath *string) error
+	// Init initializes the go-plugin client and generates a
+	// new certificate pair for gaia and the plugin/pipeline.
+	Init(command *exec.Cmd, logPath *string) error
```
Skarlso: Hmm, do you have an example of how Init and Validate would look like? Perhaps in another branch somewhere?

If I understand this correctly, you basically replaced connect with two steps. 1. Init 2. Validate?

scheduler/scheduler.go
```diff
@@ -198,9 +201,16 @@ func (s *Scheduler) prepareAndExec(r gaia.PipelineRun) {
...
 	path = filepath.Join(path, gaia.LogsFileName)
-	if err := pS.Connect(c, &path); err != nil {
+	if err := pS.Init(c, &path); err != nil {
+		gaia.Cfg.Logger.Debug("cannot initialize the plugin", "error", err.Error(), "pipeline", pipeline)
+		s.finishPipelineRun(&r, gaia.RunFailed)
+		return
+	}
+
+	// Connect to plugin(pipeline)
+	if err := pS.Validate(); err != nil {
```
Skarlso: The error message that follows this, is no longer valid. It didn't fail because it couldn't connect. It failed because the validation failed whatever that is in the plugin?

scheduler/scheduler.go
```diff
@@ -592,8 +596,14 @@ func (s *Scheduler) getPipelineJobs(p *gaia.Pipeline) ([]gaia.Job, error) {
...
+	// Init the go-plugin
+	if err := pS.Init(c, nil); err != nil {
+		gaia.Cfg.Logger.Debug("cannot initialize the pipeline", "error", err.Error(), "pipeline", p)
+		return nil, err
+	}
+
 	// Connect to plugin(pipeline)
-	if err := pS.Connect(c, nil); err != nil {
+	if err := pS.Validate(); err != nil {
 		gaia.Cfg.Logger.Debug("cannot connect to pipeline", "error", err.Error(), "pipeline", p)
```
Skarlso: This also isn't valid any longer since this is a validation error now.

scheduler/scheduler.go
michelvocks: I'm passing the job pointer to `pS.Execute` and there is set the status now: https://github.com/gaia-pipeline/gaia/pull/101/files#diff-daed06840ae2242793f6bca24ba9b8bdR185

scheduler/scheduler.go
michelvocks: Yes. The problem is `connect` did several things: 
1. Create new certificates
2. Prepares the log path/writer
3. Creates and starts the plugin
4. Validates if the plugin is compatible

It was really hard to test this one function. To somehow write valid tests for it I had to mock some things and therefore I had to split this function up into two steps. Nothing else has been changed actually.

michelvocks:
@Skarlso PHAL 🤗

Skarlso:
@michelvocks What's PHAL? :D

Either one of these?

Rank Abbr. | Meaning |  
-- | -- | --
PHAL | Public Health Association of Latvia (est. 1995) |  
PHAL | Paint Horse Association Luxembourg |  
PHAL | Portland Harbour Authority Limited (UK)

michelvocks:
Please have a look 😆

scheduler/scheduler_test.go
```diff
@@ -204,7 +212,7 @@ func TestPrepareAndExecPythonType(t *testing.T) {
 		t.Fatal(err)
 	}
 	p, r := prepareTestData()
-	pythonExecuteableName = "go"
+	pythonExecName = "go"
```
Skarlso: Shouldn't this be "python"? :D

michelvocks: No because then I have to install python/java in the test docker image used by circleci. I just fake it with the go executable. 😄

Skarlso: Argh, I see. :D

Skarlso:
LGTM! :)

Skarlso:
Dude! A failing pipeline just killed Gaia:

```bash
2018-09-06T06:37:49.007+0200 [DEBUG] plugin.failing_golang: 2018/09/06 06:37:49 CreateDeployment has been finished!
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x4 pc=0x1501680]

goroutine 1309 [running]:
github.com/gaia-pipeline/gaia/plugin.(*Plugin).Execute(0xc42063c6c0, 0xc42055c150, 0xc4204aad20, 0xc42055c150)
        /Users/hannibal/golang/src/github.com/gaia-pipeline/gaia/plugin/plugin.go:184 +0x290
github.com/gaia-pipeline/gaia/scheduler.executeJob(0x69192d8a, 0xc4200e4400, 0x12, 0xc4205de900, 0x3e, 0xc420287820, 0x1, 0x4, 0xc4200e44a0, 0x15, ...)
        /Users/hannibal/golang/src/github.com/gaia-pipeline/gaia/scheduler/scheduler.go:334 +0x151
created by github.com/gaia-pipeline/gaia/scheduler.(*Scheduler).executeScheduler
        /Users/hannibal/golang/src/github.com/gaia-pipeline/gaia/scheduler/scheduler.go:581 +0xb29
exit status 2
make: *** [dev] Error 1

```

The repo I used: https://github.com/Skarlso/go-example

michelvocks:
Thanks @Skarlso . I also found some other issues which I need to fix before we can merge that :man_facepalming:

michelvocks:
@Skarlso please have a look again. When a job throws `ErrorExitPipeline` the pipeline will be stopped but is still marked as success.

Skarlso:
@michelvocks I take a look today! Thanks for the fixes. :)

michelvocks:
@Skarlso I fixed a test so the coverage should be fine now. 🤗

Skarlso:
@michelvocks The pipeline keeps running forever after a failure. :( 

Not sure though if this is only the view or the actual job is still running... 🤔 

I think it's just the view. Also on the Log view the loader is saying the pipeline is in progress.

screen shot 2018-09-11 at 20 35 41

Validation failed is working correctly though.

Skarlso:
@michelvocks It's worse. The pipeline gets stuck pretty badly. :(

screen shot 2018-09-11 at 20 38 35

Skarlso:
@michelvocks  I fixed it if you don't mind. I found it that you didn't make the pipeline fail if there was an error. So I added it here: https://github.com/gaia-pipeline/gaia/pull/101/files#diff-daed06840ae2242793f6bca24ba9b8bdR199

Skarlso:
Could not reproduce test failure locally. :/

Skarlso:
LGTM now my man. Good to merge at your leisure @michelvocks.

michelvocks:
@Skarlso We cannot add set `FailPipeline` when an error occurs because the pipeline should not fail. It should just exit but not failing. I will have a look again :man_facepalming: :smile:

Skarlso:
@michelvocks wait. I don't get it. Why shouldn't the pipeline fail if there is an error? It shouldn't pass. If there is an error, make it fail. I don't get that reasoning? :)

michelvocks:
@Skarlso The main idea was to safely exit a pipeline. So for example:

You got a pipeline with three jobs which should be executed one after the other.
During the execution of job two you find out that job three does not need to be executed at all so you want safely exit the pipeline without failing it. 

Anyway, let me rework that. 😄

Skarlso:
But wait. I just realised what you are writing. But in my case it wasn't safely exiting. I returned an error in the job. Isn't that supposed to fail the pipeline? Or cleanup still needs to run

Skarlso:
So, I still don't understand. I found the bug, which was a timing issue with closing finish in case of a job execution failure. But this will still exit the pipeline and make it fail, because the job failed.

Also, I found that the forever running loops are pretty fragile. We need to add timeouts into those, or else we get into the trouble of an unfinished stuck pipeline which you can only kill if you kill gaia. That is unacceptable.

I'll add some timeout selects here and there to fail the loop if after a certain time there is nothing happening.

Also we might want to add user interaction like cancelling the whole thing mid run.

Skarlso:
BTW also interesting one is that I was printing out the status from the workflow after it's supposedly replaced the items, but every time on every loop it was the same output:

```bash
2018-09-19T21:57:59.149+0200 [DEBUG] plugin.def-failing_golang: 2018/09/19 21:57:59 MigrateDB has been finished!
job:  DB Migration success
job Create DB User is in state waiting for execution
job DB Migration is in state waiting for execution
job Create K8S Namespace is in state waiting for execution
job Create K8S Deployment is in state waiting for execution
job Create K8S Service is in state waiting for execution
job Create K8S Ingress is in state waiting for execution
job Clean up is in state waiting for execution
looping forever
looping forever
```

Waiting for execution. Isn't that supposed to be updating? This is Part:

```go
				for wl := range mw.Iter() {
					fmt.Printf("job %s is in state %s\n", wl.job.Title, wl.job.Status)
					if !wl.done {
						allWLDone = false
					}
				}
```

Skarlso:
Damn it. :( No, I didn't fix it. :( It made a tad bit better, but it's still hanging in an infinite loop without any channels being notified of anything.

Skarlso:
I found the problem. Fixing it..

Skarlso:
@michelvocks So the problem is the following... :)

In here `if !finalize && j.FailPipeline {` you clean up the `executeScheduler` channel of jobs still waiting to be executed and then you close the channel. So nothing will be triggered any more.

Which is fine, if a job failed, we don't want to continue execution since that would only lead to more failures if the next job depended on the previous one doing something.

However, then you go, if there are parallel jobs still running, we want them to finish and not abort, if I understand correctly. Which is again, reasonable and fine. However... Cleanup, which the next job in line, is also in started state according to the workload which only has two states, started and done.

This hangs forever, because the channel is closed and the job won't run and it's just waiting for things to happen but no channel is ever called again thus this guy:

```go
						// wait until finished
						<-notFinishedWL.finishedSig
```

blocks forever.

Skarlso:
Yiss. Fixed it.

Interesting point is that executeJob was sending one last triggerSave call which failed because we close it before hand so that's a send on closed channel.

I put the close in defer, that way it should be after that, but still prevent the select from triggering again.

Skarlso:
Nope... damn it.

```
panic: send on closed channel

goroutine 1081 [running]:
github.com/gaia-pipeline/gaia/scheduler.executeJob.func1(0xc420546840, 0xc420626000)
        /Users/gbrautigam/gohome/src/github.com/gaia-pipeline/gaia/scheduler/scheduler.go:326 +0x66
github.com/gaia-pipeline/gaia/scheduler.executeJob(0xcbe6c422, 0xc42018ef20, 0x12, 0xc42018d900, 0x3e, 0xc420010220, 0x1, 0x4, 0xc42018efc0, 0x15, ...)
        /Users/gbrautigam/gohome/src/github.com/gaia-pipeline/gaia/scheduler/scheduler.go:337 +0x143
created by github.com/gaia-pipeline/gaia/scheduler.(*Scheduler).executeScheduler
        /Users/gbrautigam/gohome/src/github.com/gaia-pipeline/gaia/scheduler/scheduler.go:582 +0xa2b
exit status 2
make: *** [dev] Error 1
```

still fails.

Skarlso:
Ideally... the writer should close the channel. But since multiple routines are using it, I'm opting that this particular part of the process handling just simply leaves this channel alone.

There is no performance drop in leaving a channel open, and it will be recycled anyways.

michelvocks:
Hey @Skarlso . I've cured my illness and now try to get back on the track. I just tried to follow all your comments but I'm a bit lost. Give me some time to get into the code again 🤗

michelvocks:
@Skarlso I found the problem. Like you said, when a job fails in a pipeline I have to wait until parallel jobs are finished.
In the go routine where I check if parallel jobs are still running I set a pointer to the current Object of an iteration object. This is not allowed because the object is always changing. I had to do a deep copy or (what I did now) just got the pointer to the actual finish channel. 

> Also, I found that the forever running loops are pretty fragile. We need to add timeouts into those, or else we get into the trouble of an unfinished stuck pipeline which you can only kill if you kill gaia. That is unacceptable.

You are absolutely right but maybe we do it in a separate (following) PR? What do you think? 🤗

michelvocks:
Ah by the way. Here is a pipeline which returns a specific error. This error tells gaia that the job failed and the pipeline should be aborted but not marked as failed: https://github.com/michelvocks/gaia-test-pipeline

Skarlso:
@michelvocks Cool stuff dude. Yeah do the timeouts in a separate PR definitely. :)

Skarlso:
I would also try to cleanup the channels later on. :)

Issue #108, opened 2018-09-21
Triggering builds in quick succession results in duplicated IDs


Skarlso:
<img width="1462" alt="screen shot 2018-09-21 at 11 46 54 am" src="https://user-images.githubusercontent.com/182850/45874074-3d890700-bd94-11e8-9f65-42230782b43c.png">

michelvocks:
Uhm. That's pretty bad. The IDs are generated by boltDB as far as I can remember. I will check that.
Thanks for the report @Skarlso  🤗

Skarlso:
@michelvocks No probs. I have a hunch that it might be only on the front-end. I didn't have the time to investigate deeper yet. That would be the best possible outcome... :) If it's deeper that's shait. We'd have to implement locking. :/

Skarlso:
Unfortunately it's boltdb.
<img width="720" alt="screen shot 2018-09-24 at 06 39 44" src="https://user-images.githubusercontent.com/182850/45937848-aa82e380-bfc4-11e8-82bb-23b3cfd8ff95.png">

Skarlso:
Aha, I see what's going on. I'll fix this.

Skarlso:
Fixed

PR #109, opened 2018-09-24
Fixed scheduling ID of pipelines

Skarlso:
Deals with #108 

screen shot 2018-09-24 at 07 16 28

### DIFF
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -243,15 +243,43 @@ func (s *Scheduler) schedule() {
 	}
 }
 
+var schedulerRunningSemaphore = make(chan bool, 0)
+var schedulerRunning = false
+
 // SchedulePipeline schedules a pipeline. We create a new schedule object
 // and save it in our store. The scheduler will later pick this up and will continue the work.
 func (s *Scheduler) SchedulePipeline(p *gaia.Pipeline, args []gaia.Argument) (*gaia.PipelineRun, error) {
+
+	// Introduce a semaphore locking here because this function can be called
+	// in parallel if multiple users happen to trigger a pipeline run at the same time.
+	// (or someone is just simply eager and presses (Start Pipeline) in quick successions).
+	// This means that one of the calls will take slightly longer (a couple of nanoseconds)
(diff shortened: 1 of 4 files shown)


michelvocks:
Your fix does not work, @Skarlso 😞 
This will increment the ID globally. Which means when you have multiple pipelines the IDs will be shared among all pipeline runs e.g. start pipeline A -> will get ID 1; start pipeline B -> will get ID 2 (should get also ID 1). 

This is also incompatible to existing boltDB databases.

Skarlso:
@michelvocks Alright. This time it's working with multiple pipelines and multiple runs with quick button smashes. It's also warding against if multiple users are using Gaia at the same time and are trying to trigger a build at the same time. :) I think you'll be happy with this one. Hopefully. :D

screen shot 2018-09-24 at 22 36 28

scheduler/scheduler.go
```diff
@@ -243,9 +243,34 @@ func (s *Scheduler) schedule() {
...
 // and save it in our store. The scheduler will later pick this up and will continue the work.
 func (s *Scheduler) SchedulePipeline(p *gaia.Pipeline, args []gaia.Argument) (*gaia.PipelineRun, error) {
+
+	// Introduce a semaphore locking here because this function can be called
+	// in parallel if multiple users happen to trigger a pipeline run at the same time.
+	// (or someone is just simply eager and presses (Start Pipeline) in quick successions).
+	// This means that one of the calls will take slightly longer (a couple of nanoseconds)
+	// while the other finishes to save the pipelinerun.
+	// This is to ensure that the highest ID for the next pipeline is calculated properly.
+	if schedulerRunning {
```
Skarlso: why did I even put an if here... 🤦‍♂️

scheduler/scheduler.go
```diff
@@ -243,9 +243,34 @@ func (s *Scheduler) schedule() {
...
+		for schedulerRunning {
```
Skarlso: Ok, I need to rethink this semaphore with a channel.

Skarlso:
Omg what am I doing. Just use a waitgroup lock unlock. That's what they are for.

PR #113, opened 2018-09-24
Fix duplicated log entries

### LINKED ISSUE #74: Multiple log entries are duplicated
screen shot 2018-08-12 at 13 02 20

When a job includes jobs which are executed in parallel then log output from these jobs are duplicated (see image).

michelvocks:
Fixes #74

### DIFF
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -30,7 +30,7 @@ const (
 	argTypeVault = "vault"
 
 	// logFlushInterval defines the interval where logs will be flushed to disk.
-	logFlushInterval = 2
+	logFlushInterval = 1
 )
 
 var (
@@ -496,7 +496,7 @@ func (s *Scheduler) executeScheduler(r *gaia.PipelineRun, pS Plugin) {
 	for {
 		select {
 		case <-finished:
-			pipelineFinished <- true
+			close(pipelineFinished)
 			return
 		case j, ok := <-triggerSave:
 			if !ok {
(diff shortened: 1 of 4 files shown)


michelvocks:
@Skarlso please have a look. This fixes the bug and also improves the general logging. Logs are now streamed to the disk every second. This improves the output a lot.

scheduler/scheduler.go
```diff
@@ -474,6 +496,7 @@ func (s *Scheduler) executeScheduler(r *gaia.PipelineRun, pS Plugin) {
 	for {
 		select {
 		case <-finished:
+			close(pipelineFinished)
```
Skarlso: You should be sending a true to this guy before closing.

Skarlso: So, you'll need to get out of the habit of doing this. :) The close doesn't actually signal the channel that it's closed or something. This only HAPPENS to work because of the side effect of reading from a closed channel. Reading from a closed channel is allowed right now, and will just not read any value, but the leg of the select will still execute. However, that's a side effect behaviour. 

The reasonable and polite thing to do to a channel if it's closed and you are the reader is to check wether you are receiving a value or you are receiving because it's closed.

For example:

```go
			select {
			case _, ok := <-done:
				if ok {
					fmt.Println("received closing signal")
					wg.Done() // do some behaviour here.
				} else {
                                         fmt.Println("the channel has been closed. terminating.")
                                 }
			default:
				fmt.Println("doing nothing")
			}
```

Why is this important? I could just put the wg.Done() into the leg and be done. Well, no. Because if that guy is in a for loop for example and there is nothing else, than that leg will run at least twice or three times before it actually terminates calling wg.Done() more often than it should and leaving wg in a negative count or less than it should be.

So please be aware of this and try not to signal a channel to finish by closing it and relying on a side effect to do the job. That side effect may be altered in future versions which would leave us unable to update without a refactor.

I hope this makes sense and doesn't sound overly picky. :) :)

michelvocks: You are absolutely right. This is not really clean and might lead to weird bugs in the future. I have updated my PR. What do you think? 🤗

Issue #110, opened 2018-09-24
Pipeline runs should be cancellable by the user


michelvocks:
A user should be able to cancel a pipeline run at any time.

PR #117, opened 2018-10-01
Cancel pipeline a pipeline run

Skarlso:


Skarlso:
Deals with #110

workers/scheduler/scheduler.go
```diff
@@ -260,6 +261,30 @@ func (s *Scheduler) schedule() {
...
+
+	// 1. Get all running Jobs
+	// 2. Set state to failed and send a finish signal
+	// 3. Store the result
+
+	pr, err := s.storeService.PipelineGetRunByPipelineIDAndID(p.ID, runID)
+	if err != nil {
+		return err
+	}
+	if pr.Status != gaia.RunRunning {
```
Skarlso: Add scheduled here as well.

workers/scheduler/scheduler.go
```diff
@@ -260,6 +261,30 @@ func (s *Scheduler) schedule() {
...
+		return errors.New("pipeline is not in running state")
+	}
+	for _, job := range pr.Jobs {
+		if job.Status == gaia.JobRunning || job.Status == gaia.JobWaitingExec {
+			job.Status = gaia.JobFailed
+			job.FailPipeline = true
```
Skarlso: This is not enough to fail a job. I need to find a way to make jobs that hang cancelled as well. Even if the process is stuck.

Skarlso:
So, this is now working, but it feels a bit wonky. Even though I tested it with multiple runs.

It will end up in cancelled state. It also works with long running processes. 

Now I have to write tests. And a bunch of them too.

What do you think so far @michelvocks? :)

screen shot 2018-10-02 at 22 49 19

michelvocks:
Good job @Skarlso 👍 

A few remarks from side (I just tested it, didn't had a look at the code yet):

1. The icon used to kill a pipeline run is a bit irritating. We should change that to something that actually mirrors the idea to kill a pipeline run. 
2. The background color of the modal panel (the confirmation) is not globally defined. If you just open Gaia and open the panel it is actually white because the CSS-Class "secret-modal" cannot be found.
3. The text in the modal panel is black which is hard to read on a dark background. 
4. The logs view is a bit broken because it's not aware of the new state which you have introduced:
screen shot 2018-10-04 at 19 01 46

Keep up the good work 👍 😄

Skarlso:
@michelvocks Thanks for the review!! :) I'll address these in haste. :)

Skarlso:
Alright, fixed the view. Going to fix the logs next.
screenshot 2018-10-05 at 21 41 00
screenshot 2018-10-05 at 21 40 54

Skarlso:
Aaaand, the log is done as well. That was easier than expected. :D

michelvocks:
Looks good to me 🤗

Skarlso:
@michelvocks Thanks. :) It's a hell of a b*tch to test the channels though. :/

michelvocks:
@Skarlso Yeah, I'm sorry. 😞 It got a bit messy. I think we have to rework this soon. Still not quite sure how...

Skarlso:
Also. The main reason this is still in WIP is this: 

```
user         13134   0.0  0.2  4481676  26808 s002  S+    9:07pm   0:04.59 /var/folders/yd/7ghbhjz521d5lk76j265_ymw0000gp/T/go-build093401996/b001/exe/main -homepath=/Users/user/golang/src/github.com/gaia-pipeline/gaia/tmp -dev=true
user         13114   0.0  0.2  4405260  31068 s002  S+    9:07pm   0:00.71 go run ./cmd/gaia/main.go -homepath=/Users/user/golang/src/github.com/gaia-pipeline/gaia/tmp -dev=true
```

The binary is still running if it's in sleep state. Even though the pipeline is cancelled. And I think that's actually not okay. So I'm going to add in some more facility which sends a SIGKILL to the binary of the pipeline.

michelvocks:
Actually as soon as `Close` is called, the plugin process should be killed by go-plugin. It's important to know that the close of the plugin runs asynchronus and it might take a while till the plugin gets killed.

How long did you wait?

Skarlso:
It did not kill it. I opened a ticket on go-plugin. I'll write down some more details on Sunday, but I'm out right now until then.

Skarlso:
Back from vacation. Going to investigate why the plugin did not get killed.

Skarlso:
Alright, let's see... I'm running the pipeline and waiting for a while for go-plugin to shut it down.

```
hannibal         47896   0.0  0.1  4403772  23448 s002  S+    1:40pm   0:01.95 /var/folders/yd/7ghbhjz521d5lk76j265_ymw0000gp/T/go-build174948883/b001/exe/main -homepath=/Users/hannibal/golang/src/github.com/gaia-pipeline/gaia/tmp -dev=true
hannibal         47857   0.0  0.2  4424240  28600 s002  S+    1:40pm   0:00.98 go run ./cmd/gaia/main.go -homepath=/Users/hannibal/golang/src/github.com/gaia-pipeline/gaia/tmp -dev=true
```

Skarlso:
@michelvocks It's not killing the plugin. It's running for 11 minutes now, and nothing's happening with it.

The job itself is sleeping. so I think while the healthcheck must be succeeding the job it self is still stuck and thus it doesn't get shut down.

I'm thinking of trying to locate the running process based on the name of the execution binary. Once I have a PID I'll just send a signal to it.

Skarlso:
@michelvocks NEVER MIND! I'm blind. I had a hunch I was looking at the wrong process. Turns out I was. :D

```
❯ ps aux |grep long-sleep
hannibal         57628   0.0  0.1  4396904  10384 s002  S+    1:57pm   0:00.02 /Users/hannibal/golang/src/github.com/gaia-pipeline/gaia/tmp/pipelines/long-sleep_golang
```

And this guy is dead before it even hit the floor. After cancel it successfully disappeared.

While I can't really unit test the channels I think this is as done as it gets.

PR #121, opened 2018-10-05
Added a definitive check for possible password errors.

Skarlso:
This deals with #120.

Instead of doing this bit: https://github.com/gaia-pipeline/gaia/blob/413494fa79a632c4dace349aebfdb5009eb5bd54/security/vault.go#L256-L261

Now there is a definitive way of checking if the decryption was successful or not.

### DIFF
--- a/security/vault.go
+++ b/security/vault.go
@@ -18,7 +18,9 @@ import (
 )
 
 const (
-	vaultName = ".gaia_vault"
+	vaultName        = ".gaia_vault"
+	secretCheckKey   = "GAIA_CHECK_SECRET"
+	secretCheckValue = "!CHECK_ME!"
 )
 
 // VaultAPI defines a set of apis that a Vault must provide in order to be a Gaia Vault.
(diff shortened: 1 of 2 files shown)


security/vault.go
```diff
@@ -190,6 +190,7 @@ func (v *Vault) encrypt(data []byte) (string, error) {
 		// User has deleted all the secrets. the file will be empty.
 		return "", nil
 	}
+	data = append(data, []byte("\nGAIA_CHECK=!CHECK_ME!")...)
```
michelvocks: That looks a bit magic to me. What do you think about moving this to constant variable?

security/vault.go
```diff
@@ -241,6 +242,10 @@ func (v *Vault) decrypt(data []byte) ([]byte, error) {
 	if err != nil {
 		return []byte{}, err
 	}
+
+	if !bytes.Contains(unpadMsg, []byte("!CHECK_ME!")) {
```
michelvocks: Same here. A bit magic. Maybe move that to a separate constant with a tiny description? 😄

security/vault.go
```diff
@@ -253,11 +258,8 @@ func (v *Vault) parseToMap(data []byte) error {
 	row := bytes.Split(data, []byte("\n"))
 	for _, r := range row {
 		d := bytes.Split(r, []byte("="))
-		if len(d) < 2 {
-			// It is possible that if there is a password failure it's not caught
-			// by the padding process. Here it will be caught because we can't
-			// marshal the data into proper k/v pairs.
-			return errors.New("possible mistyped password")
+		if bytes.Equal(d[0], []byte("GAIA_CHECK")) {
```
michelvocks: Same like above

security/vault.go
Skarlso: Sure thing! 😊

PR #123, opened 2018-10-09
Add job "running" icon to pipeline run graph

### LINKED ISSUE #112: Add job "running" icon to pipeline run graph
The detailed view of a pipeline run includes the job graph with shows the dependencies between the jobs and the current state. 

Currently, three states are implemented: Waiting for execution, Success, Failed.
To improve the user experience we should also add a new icon (and state) for Running.
It would be cool to use an animated gif for this state so it makes it easy for the user to see the current active job.

michelvocks:
Fixes #112


Skarlso:
Stuff's working man. Nice. Looks a bit weird though. What's up with the Red border?

screen shot 2018-10-10 at 6 15 27 pm

Issue #73, opened 2018-08-12
Fix scale of pipeline views in overview on large screens


michelvocks:
![screen shot 2018-08-12 at 13 00 06](https://user-images.githubusercontent.com/2463119/44001305-bc52c55a-9e2f-11e8-8a80-2f3b5aa1c376.png)

On large screens, there is too much space between multiple pipelines.

speza:
Thought I'd quickly give this one a fix. Few little tweaks but seems much better locally for me - 
![capture 2018-10-16 at 17 45 30](https://user-images.githubusercontent.com/5376774/47033057-cadd3280-d16b-11e8-8e87-147b75722ec2.png)

PR #128, opened 2018-10-16
Fix scaling of pipelines on overview (#73)

speza:


Skarlso:
Hi! First of all, thanks for this contribution! 

Second, this now anchors the divs to each other. Making them stick, so when the screen is large, they get elongated. Meaning the span and widen. 

I honestly think that that doesn't look so good. A better option is to make the divs re-arrange on larger screens so they take up the place, but do that without changing the size of the dive, but re-arranging themselves.

This is how it looks like now on wider view mode:

screenshot 2018-10-16 at 21 53 29

What do you think @michelvocks?

Skarlso:
I'm by no means good at front-end stuff, but what I'm talking about is basically and auto-balancing css grid. There are a couple of articles about it around, if you don't happen to know that stuff, it's a great learning opportunity. :) Good luck! 👍

speza:
I am also by no means a front-end guy! Was just looking for a ‘quick win’ while in some downtime. I just looked for a quick solution using the UI framework already in-place. Probably a good thing to know so I will look into a better solution soon.

speza:
@Skarlso let me know what you think of this latest change.

Skarlso:
@speza excellent!! It works lovely. :)

michelvocks:
@speza Thanks a lot for your contribution. This is really cool! ❤️

PR #132, opened 2018-12-06
Python create pipeline race condition

### LINKED ISSUE #131: Python create pipeline race condition
I noticed that the update process during the pipeline creation process has a race condition. This results in failed pipeline builds sometimes.

michelvocks:
Fixes #131

### DIFF
--- a/frontend/client/views/pipeline/create.vue
+++ b/frontend/client/views/pipeline/create.vue
@@ -371,12 +371,12 @@ export default {
 
                 if (this.historyRows[i].status >= 50) {
                   this.historyRows[i].output += 'Pipeline has been successfully compiled.\n'
-                  this.historyRows[i].output += 'Copy binary to pipelines folder...\n'
+                  this.historyRows[i].output += 'Verifying pipeline...\n'
                 }
 
                 if (this.historyRows[i].status >= 75) {
-                  this.historyRows[i].output += 'Pipeline binary has been copied to pipelines folder.\n'
-                  this.historyRows[i].output += 'Starting integrity check...\n'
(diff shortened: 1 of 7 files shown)


michelvocks:
The main problem was a race condition between our ticker (https://github.com/gaia-pipeline/gaia/blob/master/workers/pipeline/ticker.go#L67) and the create_pipeline process (https://github.com/gaia-pipeline/gaia/blob/master/workers/pipeline/create_pipeline.go#L28). 

It's important that during the create_pipeline process the ticker will not interfere. That was the case because currently we have this workflow:

1. Clone git repo
2. Build pipeline
3. Copy pipeline into pipeline folder <--- ticker will now pick up the pipeline!
4. Validate pipeline (e.g. start pipeline and get all jobs)
5. Save pipeline data into store. 

Usually, this was not a problem except for python because the build process is a bit more complex here (build archive, setup virtual python environment etc.).

Anyway, in my opinion this solution now is more determistic and fixes the race condition:

1. Clone git repo
2. Build pipelne
3. Validate pipeline
4. Save pipeline into store
5. Copy Binary into pipelines folder <-- ticker can pick it up but we finished anyway.

michelvocks:
Ping @Skarlso 😄 
Sorry, I know you are busy but I need this for my other PR. Would be awesome if you could just have a quick look at it. 🤗

Skarlso:
Damn! Totally meant to do it! I even have a calendar event. Just got absolutely swamped with work. Sorry bro.

michelvocks:
@Skarlso No problem 🤗  Let me know when you got some free time 😄

PR #133, opened 2018-12-07
Acceptance/Integration Tests introduction

michelvocks:
Acceptance/Integration tests are really important because they ensure that multiple components can still work together. In the past, we did this manually which had a negativ impact on our feature development time as well as it opened the door to introduce misterious side effects into Gaia.

This PR is the start of implementing an Acceptance/Integration test framework (we just use go tests) which also requires some refactoring.


michelvocks:
One job is failing because of the race condition which should be fixed with my other PR. Once #132 has been merged and the job doesn't crash it should be ready to get merged! 🤗

PR #138, opened 2018-12-14
Added/Updated screenshots for README

michelvocks:
I updated the screenshots and instead of using rawgit (which is deprecated) we are now using the same repository.

Currently, the README in this PR is broken and does not show any screenshots because the link points to gaia-pipeline/gaia/blob/master. Once this PR has been merged it will work as usual.


michelvocks:
Thanks @Skarlso 🤗 
Yeah, codecov is wierd sometimes 🤔

PR #139, opened 2018-12-21
Implement pipeline scheduling via cron syntax

### LINKED ISSUE #105: Implement pipeline scheduling via cron syntax
A good library to start with could be: https://godoc.org/github.com/robfig/cron

michelvocks:
fixes #105

### DIFF
--- a/handlers/pipeline.go
+++ b/handlers/pipeline.go
@@ -10,6 +10,7 @@ import (
 	"github.com/gaia-pipeline/gaia/services"
 	"github.com/gaia-pipeline/gaia/workers/pipeline"
 	"github.com/labstack/echo"
+	"github.com/robfig/cron"
 	uuid "github.com/satori/go.uuid"
 )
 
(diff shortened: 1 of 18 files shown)


michelvocks:
This is ready to be merged 🤗

Skarlso:
@michelvocks Well done! I'll take a look as soon as I can. :)

handlers/pipeline.go
```diff
@@ -183,12 +184,66 @@ func PipelineUpdate(c echo.Context) error {
...
+		// Update active pipelines
+		pipeline.GlobalActivePipelines.Replace(foundPipeline)
 	}
 
 	return c.String(http.StatusOK, "Pipeline has been updated")
 }
 
+// stringSliceEqual is a small helper function
+// which determines if two string slices are equal.
+func stringSliceEqual(a, b []string) bool {
```
Skarlso: Why not use reflect.DeepEqual here?

handlers/pipeline.go
```diff
@@ -183,12 +184,66 @@ func PipelineUpdate(c echo.Context) error {
...
+			foundPipeline.CronInst.Stop()
+		}
+		foundPipeline.CronInst = cron.New()
+
+		// Iterate over all cron schedules.
+		for _, cron := range p.PeriodicSchedules {
+			err := foundPipeline.CronInst.AddFunc(cron, func() {
+				_, err := schedulerService.SchedulePipeline(&foundPipeline, []gaia.Argument{})
+				if err != nil {
+					gaia.Cfg.Logger.Error("cannot schedule pipeline from periodic schedule", "error", err, "pipeline", foundPipeline)
```
Skarlso: Why not return the error here?

handlers/pipeline.go
michelvocks: Good point. I've learned that `reflect.DeepEqual` is super slow and you should only use that for tests (e.g. https://programming.guide/go/compare-slices.html). Would love to hear your opinion 😉

handlers/pipeline.go
michelvocks: Because `AddFunc` only accepts functions without return values. This is executed in it's own go-routine.

handlers/pipeline.go
Skarlso: It depends on one often do you think this function is being called versus the complexity that it adds in terms of things that you need to cover by tests. :) If you are happy with it, it's okay. :)

PR #146, opened 2019-01-07
[WIP] Make Gaia go modules compatible

Skarlso:
Testing Circle CI.


Skarlso:
That's odd. It's using the right version of uuid. The module is saying 1.2.0 which was in the dep package definition as well.

Skarlso:
Fun fact... locally it works.

fubarhouse:
@Skarlso,

Try `go mod download`.

Skarlso:
`/Users/gbrautigam/gohome/pkg/mod/github.com/satori/go.uuid@v1.2.0/generator.go`
Even using the right one.... 

@fubarhouse -- thanks, I'll try that. Although `go get` with Environment enabled should do the say I think. not?

Skarlso:
What even, circleci? 
```
go: unknown environment setting GO111MODULE=ON
```

Skarlso:
F*CKING circleci documentation. :D It says: `GO111MODULE=ON`.

Where actually go only allows for `GO111MODULE=on` case sensitive.

fubarhouse:
An example of something I'm loosely working on (to the same effect):
https://github.com/fubarhouse/ansible-role-tester/blob/issue/94/.travis.yml

``` yaml
  # Install golang deps
  - GO111MODULE=on go mod download
  - GO111MODULE=on go install .
  # Style checks
  - diff -u <(echo -n) <(gofmt -d $(find . -not -path "./vendor/*" -name "*.go"))
- GO111MODULE=on go vet $(go list ./...)
```

Skarlso:
@michelvocks hey man. I think all works fine now, but you may need to update the Gaia container if we move ahead with this.

fubarhouse:
> @Skarlso Cool. Will do that later. Just wondering if we could do the same for the go sdk? It should be backwards compatible, right? (e.g. go get still works)

`go get` does still work. `GO111MODULE` will be set to `on` as of Go 1.12 and `go get` will be removed from the toolchain for Go 1.13 - early next year.

Skarlso:
@michelvocks  It should be backwards compatible yeah.

@fubarhouse Curious. Where did they say they will remove get from the toolchain? I saw a massive change movement into the direction of modules and a decentralised module repo but not the remove of go get.

fubarhouse:
It's implied here:
```
Our aim is for Go 1.13, scheduled for August 2019, to enable module mode by default (that is, to change the default from auto to on) and deprecate GOPATH mode. In order to do that, we’ve been working on better tooling support along with better support for the open-source module ecosystem. 
```

Source: https://blog.golang.org/modules2019 

It doesn't say it'll be deleted, but it does explicitly state it will be deprecated and I have read it will be removed as of Go1.13 - though that may get pushed back for some time when push comes to shove.

docker/Dockerfile
```diff
@@ -15,7 +15,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 		pkg-config \
 	&& rm -rf /var/lib/apt/lists/*
 
-ENV GOLANG_VERSION 1.10.3
+ENV GOLANG_VERSION 1.11
```
michelvocks: Does that work? Below it will also check the md5 hash which has changed. Have a look at the other dockerfile change: https://github.com/gaia-pipeline/gaia/pull/146/files#diff-192af0c41f653dbf5bf864ba3ac4c682

Skarlso: Uh, nice. I wonder if that's the problem. :D

Skarlso:
@fubarhouse I think you're mistaken in that regard. There are no plans to deprecate or remove `get` from the tooling. I talked with a couple other people as well on IRC and Slack, all contributors, and know nothing about that. :)

fubarhouse:
@Skarlso,

Good to know for reference, thanks. 👍

Skarlso:
@michelvocks Thanks dude. :)

Skarlso:
@michelvocks do you think there is anything else with this? 🤔

docker/Dockerfile
Skarlso: Oh right, I totally thought it is by the changes of the hashes in the above. I'll fix these. :)

Skarlso:
huh. I wonder if that is just flaky... In which case we should probably refactor it.

Skarlso:
```
2019-01-07T15:09:59.759Z [DEBUG] plugin: plugin process exited: path=/bin/echo
--- FAIL: TestInit (0.00s)
    plugin_test.go:81: Error should contain 'Unrecognized remote plugin message' but was 'plugin exited before we could connect
        
        --- output ---
        <nil>'
```

Skarlso:
Interesting. Passes locally.

Skarlso:
@michelvocks I think it's ready now. :)

PR #158, opened 2019-01-27
[WIP] Remote triggering a pipeline

Skarlso:
So for now a couple of things:

- [x] The `auto` user needs to be undeletable like the admin user. Still needs to be modifiable because the token might need to be updated
- [x]  View/Edit the token for a pipeline
- [x] Write Tests!!
- [x]  Make the end-point more secure. Right now I'm not checking if the trigger was by the auto user with the correct auto token
- [x]  Deal with Pipeline parameters

Manual:

```
~/goprojects/gaia
❯ curl -X POST http://127.0.0.1:8080/api/v1/pipeline/1/4565632f-a77b-551b-b0cb-67e74dc9b15a/trigger
Trigger successful for pipeline: 1
```


Skarlso:
Alright, triggering works:

```
❯ curl -X POST -H "Authorization: Basic YXV0bzo0YzE4YTkyNi1kNWE2LTVlYTMtODY5OC1jNmQ4NDllZGFiMDE=" http://127.0.0.1:8080/api/v1/pipeline/3/9e77e732-b0ce-54d8-a2c7-c45b9c99f9c0/trigger
Trigger successful for pipeline: 3
```

This made the above pipeline execute:

screenshot 2019-02-01 at 15 54 11

Alright, now for arguments:

❯ curl -X POST -H "Authorization: Basic YXV0bzo0YzE4YTkyNi1kNWE2LTVlYTMtODY5OC1jNmQ4NDllZGFiMDE=" -H "content-type:application/json" -d '{"args": [{"type": "String","key": "username","value": "ASDFUSER"}]}' http://127.0.0.1:8080/api/v1/pipeline/5/c1741ea3-2d28-53e1-9b17-a002b8673206/trigger
Trigger successful for pipeline: 5

Skarlso:
Alright... was sending wrong json data. :/ To my defense the error from echo wasn't really helpful:

```
 curl -X POST -H "Authorization: Basic YXV0bzo0YzE4YTkyNi1kNWE2LTVlYTMtODY5OC1jNmQ4NDllZGFiMDE=" -H "content-type:application/json" -d '[{"type": "String","key": "username","value": "ASDFUSER"}]' http://127.0.0.1:8080/api/v1/pipeline/5/c1741ea3-2d28-53e1-9b17-a002b8673206/trigger
```

Skarlso:
@michelvocks How do you like the display of the tokens?
screenshot 2019-02-02 at 11 48 26
screenshot 2019-02-02 at 11 40 43

Skarlso:
WTH. Somehow I messed up the handlers? :O

Skarlso:
Oh, I just realised I don't need a new window... I can slide the user password thingy. I don't need a new icon.

Skarlso:
Ah, I'll only allow token reset for auto user. Since the reset saves the user with an empty password.

michelvocks:
@Skarlso I'm not sure if the overview page is the right place for the token display. This information is usually needed once (or twice) but it's not a piece of information which you need regularly. 

In my opinion, we need an information view in the pipeline details page. Currently, when you open the details page for pipeline you basically see no information about the pipeline (e.g. Name of the pipeline, last successful run, last failed run etc.). 

Regarding the Auto User: Not quite sure if we should store for every user a UUID when you can only use it for one specific user. And the Auto User needs to be created by the user? Why we are not creating it automatically?

Skarlso:
@MichelVocks I agree on the pipeline view. It's not the best place.

I am generating the auto user. 🙂 Also agreed that that should be the only user with a token. Hence I'm restricting resetting tokens for any other user. Also I'm disallowing deleting this user. I haven't implemented that yet.

frontend/client/views/settings/index.vue
```diff
@@ -260,6 +291,37 @@
...
 
+    <!-- reset trigger token modal -->
+   <modal :visible="showResetPipelineTriggerTokenModal" class="modal-z-index" @close="close">
+      <div class="box pipeline-modal">
+        <div class="block pipeline-modal-content">
+          <collapse accordion is-fullwidth>
+            <collapse-item title="Reset Pipeline Trigger Token" selected>
+              <div class="pipeline-modal-content">
+                <p class="control has-icons-left" style="padding-bottom: 5px;">
+                  <!-- <input class="input is-medium input-bar" v-focus v-model="selectPipeline.name"
```
michelvocks: If it's not needed we should remove it. Later nobody will remember why this was there. 😄

frontend/client/views/settings/index.vue
```diff
@@ -483,6 +563,22 @@
         this.close()
       },
 
+      resetUserTriggerToken () {
+        this.$http
+          .get('/api/v1/user/' + this.selectUser.username + '/reset-trigger-token')
```
michelvocks: In my opinion, it should be PUT/POST/PATCH since this is an update operation?

frontend/client/views/settings/index.vue
```diff
@@ -582,6 +678,22 @@
         this.close()
       },
 
+      resetPipelineTriggerToken () {
+        this.$http
+          .get('/api/v1/pipeline/' + this.selectPipeline.id + '/reset-trigger-token')
```
michelvocks: In my opinion, it should be PUT/POST/PATCH since this is an update operation?

workers/pipeline/create_pipeline.go
```diff
@@ -139,6 +141,16 @@ func CreatePipeline(p *gaia.CreatePipeline) {
 		return
 	}
 
+	nsUUID := uuid.NewV4()
```
michelvocks: In my opinion, this should be moved to https://github.com/gaia-pipeline/gaia/pull/158/files#diff-3ea375c9ea9839dd14147d6613685a41R127

It's risky to save the pipeline object after "CopyBinary" because that introduces a race condition with the ticker.

Skarlso: Woa. Nice catch! Thanks.

frontend/client/views/settings/index.vue
Skarlso: Hah, thanks!

frontend/client/views/settings/index.vue
Skarlso: Agreed. Get is not very nice. :)

workers/pipeline/create_pipeline.go
Skarlso: @michelvocks Where does your link point to? 🤔

Skarlso: @michelvocks Like that? :)

workers/pipeline/create_pipeline.go
```diff
@@ -62,6 +62,16 @@ func CreatePipeline(p *gaia.CreatePipeline) {
...
```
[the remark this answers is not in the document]
Skarlso: 👍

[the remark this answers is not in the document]
Skarlso: Oh Gods. Why did I even do this? I copied the whole save. When I had a save there? WHY? WHY BRAIN?

Skarlso:
@michelvocks Np! Thanks! :)

workers/pipeline/create_pipeline.go
```diff
@@ -124,7 +125,11 @@ func CreatePipeline(p *gaia.CreatePipeline) {
 	}
 
 	nsUUID := uuid.NewV4()
-	p.Pipeline.TriggerToken = uuid.NewV5(nsUUID, "pipelineTriggerToken").String()
```
Skarlso: @michelvocks I just realised that this was hardcoded namespace. Although the previous UUID was pseudo-random, hardcoding the namespace here is like giving it a seed of some sort. It was probably fine because of the random UUID but... you never know. :) Unless I would have wanted to create a hierarchy, I just feel I will sleep better knowing that it's not. :D

michelvocks: @Skarlso Thanks for pointing that out. I'm currently working on the secret generation for the worker and I would like to use the same functionality you use here. Maybe we should refactor that to an external place? Maybe `security`?

Skarlso: @michelvocks Hmm, okay, sure. That's a good suggestion. I'll do it.

Skarlso:
@michelvocks Done. What do you think?

michelvocks:
Perfect, thanks @Skarlso 
Feel free to merge it.

PR #159, opened 2019-01-28
Fixed checksum mismatch

michelvocks:
Our build is currently broken because of a checksum mismatch from `echo`:

```
go mod download
go: verifying github.com/labstack/echo@v3.3.5+incompatible: checksum mismatch
	downloaded: h1:Y3vG4kINVWNQN8Y6Jdur8uLat7fSLV5n5yLE8n+JbF4=
	go.sum:     h1:9PfxPUmasKzeJor9uQTaXLT6WUG/r+vSTmvXxvv3JO4=
Makefile:31: recipe for target 'download' failed
make: *** [download] Error 1
Exited with code 2
```

I was able to fix it by manually removing the dependency and add it by running `go mod tidy`. This added also new indirect dependencies which is fine in my opinion. Not quite sure what's the exact cause of this mismatch is. Any ideas, @Skarlso ?


Skarlso:
@michelvocks Me neither. I was also researching this with no avail. Maybe something went wrong with the build process, or echo released something or change something.. I'm not sure.: /

Issue #46, opened 2018-07-20
[Feature Request] Distributed execution


vosmith:
Do you have plans to use turn Gaia into a Master <-> Worker system where jobs can be run on remote servers?

michelvocks:
Welcome @vosmith 🤗 
Yes, this feature is planned. Sadly, I cannot tell you when we are going to implement this. 
Our feature list is long and there are other features more important than this. 😄

michelvocks:
Now supported via #166

PR #166, opened 2019-02-02
Distributed execution via worker

### LINKED ISSUE #107: RFC: Distributer Pipeline execution via Workers
# Abstract

This document discusses the problem of executing pipelines in a distributed
manner.

# Table of Contents

1. [Introduction](#introduction)
2. [Problem Statement](#problem-statement)
3. [Terminology](#terminology)
4. [Architecture Diagram](#architercutre-diagram)
5. [Proposed Worker Distribution Model](#proposed-worker-distribution-model)
6. [Managing Workers](#managing-workers)
7. [Worker Tags](#worker-tags)
8. [The Worker RPC API](#the-worker-rpc-api)
9. [Gaia Master - Agent](#gaia-master-agent)
10. [Scheduling Jobs](#scheduling-jobs)
11. [Implementation Approach](#implementation-approach)

# Introduction

# Problem Statement

The problem poses the following set of challenges for Gaia:

1. Manage workers
    - See what pipeline is running on which worker at any given point in time
    - Add / Delete / Suspend workers
    - Add specific environment variables to the worker
2. Either automatically, or manually choose which pipeline should run on which
worker.
3. Label the workers so the user knows it's a windows machine or a linux machine
or Go, Python, Java SDK is available on it... etc.

# Terminology

**Gaia Master**: The Gaia Master is a running instance of gaia launched via make or the
released Gaia binary.
**Worker**: A worker is a server which is connected to the Gaia Master and has
certain capabilities like, what kind of SDK it supports or what operating system
is installed on it.
**Pipeline**: A pipeline is a configured entity with a set of Jobs.
**Job**: A job is a single running task like, create a user. A pipeline can have multiple jobs.
**RPC**: Remote Procedure Call

# Architecture Diagram

distributed workers

# Proposed Worker Distribution Model

The proposed model which aims to solve this problem is laid out as follows. 

# Managing Workers

The managing of the workers will happen through a set of API endpoints.
All workers are stored in the database with a designated set of labels
assigned name and IP address.

These endpoints will be Delete / Suspend. Since adding will be taken care of
by the Gaia agent, we don't support that operation here specifically.

**Delete**: Delete will simply remove the server from the rotation. It won't restart
the server, or shut it down, it will just simply delete it from the database which
holds the worker instances.
**Suspend**: Suspending a worker will take it out of rotation but will not delete it.
Suspended this worker will not be able to run any pipelines. This is a good option
if some kind of maintenance needs to be performed on the machine.

## Worker Tags

The workers will need to be tagged with what kind of resource they are providing. For example:

| name  | tags |
|-------|------|
| Worker 1 | Ubuntu Linux 64bit |
| Worker 2 | Windows 10 64bit |
| Worker 3 | Debian Linux 64bit |

When a pipeline is first created in needs to set on the pipeline creation window what kind of resources it requires. These tags will need to be made accessible by a drop down list for ease of usage. These tags can be created when a Worker is created and saved to Gaia. Tagging them can be done manually on the Worker Manager screen.

## The Worker RPC API

The Workers will talk to the Gaia Master via a set of defined RPC interfaces.
These are as follows:

```go
// RegisterWorker will take a worker struct which contains the following information:
// Security: This will be protected by the TLS connection between master and worker.
// IP: The address of the worker
// Name: The name of the worker which typically can be `hostname`.
// Operating System: The OS of the worker to save as a label.
// SDK: The SDK the worker has.
rpc RegisterWorker(Worker) {}

// RunPipeline will take a pipeline, and execute it. This ia bi-directional endpoint.
// Pipeline struct:
// ID: Id of the pipeline
// Repo: The git repository for the pipeline. This is needed because the worker needs
// to build the pipeline.
rpc RunPipeline(Pipeline) returns (Success) {}

rpc GetAllPipelines(Worker) returns (Pipelines) {}
```

# Gaia Master - Agent

The current Gaia implementation will still hold and will be designated as Gaia Master.
The master will be a hub for the worker to connect to, get pipelines from, and report
back on the current state of the pipelines they are running.

As such, Gaia Master will no longer be solely responsible to build and distribute
binaries. Since the operation system of the worker decides in what format the binaries
will be in, the workers will build their own binaries.

Which means a worker will get a repository to pull code from and do the whole thing
that Gaia does currently. This will not involve duplicating code however, since the
whole thing will be in the `worker` package. Gaia Master will use this package by
setting worker to `localhost`.

The Workers will need to have the go-plugin extracted because HashiCorp's plugin
system does not support RPC calls over the network. Just strictly localhost communication
is allowed. Pipeline execution and communication between jobs' running and state
changes are all through RPC.

# Scheduling Jobs

Scheduling jobs will also have to be included into the workers. Workers will schedule
their own parallel jobs execution model and Gaia Master will have to schedule and manage
which worker to distribute pipelines to. This means that the workers will need an indicator
to define when they are too busy to accept more pipelines.

# Where jobs are built

Currently, once a user initiates a pipeline build, that pipeline is saved and built on Gaia Master.
This has to change in order for the worker to be able to run the pipeline. The binary
needs to be built on the worker. However, Gaia also needs to be aware of the jobs,
and does pre-validation which means it also needs to build the pipeline.

## Scenario 1: 
We build the pipeline on both, the Gaia master, and the Worker. Which means we get immediate validation of the pipeline but have to duplicate the building process.

## Scenario 2:
We only build on the worker and just save the pipeline on the master to track it. The validation will be deferred until it's actually built on one of the workers. This way, validation is deferred but the building process isn't duplicated.

# Implementation Approach

1. Extract all functionality regarding running and building pipelines including
the SDK and the go-plugin facility into a worker package. This should not change
the current behavior of Gaia. All tests should still pass. Including the WebHook
capability which should be able to still just call build. The worker package should
take care of building and distributing the binary.

2. Create the API which handles most of the things worker related. But still don't
bother extracting it.

3. Create an Agent binary which calls back to master's RPC API and registers a
server as a worker.

4. Implement the managing of the servers below settings on the left of the admin
screen.

michelvocks:
Fixes #107 and #46 

Related docs PR: https://github.com/gaia-pipeline/docs/pull/11

TODO:
- [x] Add tags to worker UI
- [x] Fix tab panel order in UI
- [x] Add option to prevent primary instance to accept work
- [x] Write more tests
- [x] Write documentation


server/server.go
```diff
@@ -146,25 +138,66 @@ func Start() (err error) {
...
+			gaia.Cfg.VaultPath = gaia.Cfg.DataPath
+		}
+		v, err := services.VaultService(nil)
+		if err != nil {
+			gaia.Cfg.Logger.Error("error initiating vault")
+			return err
+		}
+
+		// Generate global worker secret if it does not exist
+		secret, err := v.Get(gaia.WorkerRegisterKey)
```
golangcibot: ineffectual assignment to `secret` (from `ineffassign`)

mohammadchehab (approved):

workers/agent/agent.go
```diff
@@ -0,0 +1,583 @@
...
+		}
+
+		// Get pipeline binary name and SHA256SUM
+		pipelineName := pipelineRunPB.PipelineName
+		pipelineSHA256SUM := pipelineRunPB.ShaSum
+
+		// Setup reschedule of pipeline in case something goes wrong
+		reschedulePipeline := func() {
+			pipelineRunPB.Status = string(gaia.RunNotScheduled)
+			a.client.UpdateWork(context.Background(), pipelineRunPB)
```
golangcibot: Error return value of `a.client.UpdateWork` is not checked (from `errcheck`)

workers/server/worker.go
```diff
@@ -0,0 +1,232 @@
...
+				break
+			}
+		}
+
+		// Stream pipeline run back to worker
+		if err = serv.Send(&gRPCPipelineRun); err != nil {
+			gaia.Cfg.Logger.Error("failed to stream pipeline run to worker instance", "error", err.Error(), "worker", workInst)
+
+			// Insert pipeline run back into memdb since we have popped it
+			db.InsertPipelineRun(scheduled)
```
golangcibot: Error return value of `db.InsertPipelineRun` is not checked (from `errcheck`)

workers/server/worker.go
```diff
@@ -0,0 +1,232 @@
...
+
+		// Get memdb service
+		db, err := services.MemDBService(nil)
+		if err != nil {
+			gaia.Cfg.Logger.Error("failed to get memdb service via updatework", "error", err.Error())
+			return e, err
+		}
+
+		// Put pipeline run back into memdb. This adds the pipeline run to the stack again.
+		db.InsertPipelineRun(run)
```
golangcibot: Error return value of `db.InsertPipelineRun` is not checked (from `errcheck`)

workers/agent/agent.go
```diff
@@ -0,0 +1,637 @@
...
+				// Set bytes
+				chunk.Chunk = buffer[:bytesread]
+
+				// Stream it to primary instance
+				if err = stream.Send(chunk); err != nil {
+					gaia.Cfg.Logger.Error("failed to stream log chunk to primary instance", "error", err.Error(), "pipelinerun", run)
+					return
+				}
+			}
+			stream.CloseSend()
```
golangcibot: Error return value of `stream.CloseSend` is not checked (from `errcheck`)

Skarlso:
Duuddee! Nice, I'm going to have to review this at one point. :D

Skarlso:
BTW, this should have been a bit separated into multiple PRs if possible. :D Like, the worker view, then the pipeline stuff and so on and so forth. :)

michelvocks:
@Skarlso I thought about splitting this PR up in multiple PRs beforehand but this almost impossible. The way this has been implemented now is a bit different from what we agreed on in the RFC since some things didn't work well or included some traps we didn't think of. 😭 

Let me check if I can split this up when it's finished. 😄

Skarlso:
@michelvocks It's fine, it will just be harder to review. :D 

Ah, would you mind updating the RFC in that case? :P

handlers/workers_test.go
```diff
@@ -0,0 +1,159 @@
...
+	// Generate global worker secret
+	secret := []byte(security.GenerateRandomUUIDV5())
+	v.Add(gaia.WorkerRegisterKey, secret)
+	if err := v.SaveSecrets(); err != nil {
+		t.Fatal(err)
+	}
+
+	// Initialize echo
+	e := echo.New()
+	InitHandlers(e)
```
golangcibot: Error return value of `InitHandlers` is not checked (from `errcheck`)

handlers/workers_test.go
```diff
@@ -0,0 +1,159 @@
...
+		body := registerWorker{
+			Secret: "random-wrong-secret",
+		}
+		bodyBytes, _ := json.Marshal(body)
+		req := httptest.NewRequest(echo.POST, "/api/"+gaia.APIVersion+"/worker/register", bytes.NewBuffer(bodyBytes))
+		req.Header.Set("Content-Type", "application/json")
+		rec := httptest.NewRecorder()
+		c := e.NewContext(req, rec)
+
+		RegisterWorker(c)
```
golangcibot: Error return value of `RegisterWorker` is not checked (from `errcheck`)

handlers/workers_test.go
```diff
@@ -0,0 +1,159 @@
...
+			Secret: string(secret[:]),
+			Tags:   []string{"first-tag", "second-tag", "third-tag"},
+		}
+		bodyBytes, _ := json.Marshal(body)
+		req := httptest.NewRequest(echo.POST, "/api/"+gaia.APIVersion+"/worker/register", bytes.NewBuffer(bodyBytes))
+		req.Header.Set("Content-Type", "application/json")
+		rec := httptest.NewRecorder()
+		c := e.NewContext(req, rec)
+
+		RegisterWorker(c)
```
golangcibot: Error return value of `RegisterWorker` is not checked (from `errcheck`)

services/service_provider_test.go
```diff
@@ -48,10 +48,13 @@ func TestSchedulerService(t *testing.T) {
 	if schedulerService != nil {
 		t.Fatal("initial service should be nil. was: ", schedulerService)
 	}
-	SchedulerService()
-	defer func() { schedulerService = nil }()
-	if schedulerService == nil {
-		t.Fatal("service should not be nil")
+	StorageService()
```
golangcibot: Error return value of `StorageService` is not checked (from `errcheck`)

michelvocks:
@Skarlso This is ready for review. Need to work on a few tests and documentation but overall ready for initial review.

Skarlso:
wooot!! Holy Moly! :D I'll test it out... The code review is massive. :D So that will take some time. :P

Skarlso:
`+5,700 −954`

Nooooooo. :D

michelvocks:
@Skarlso Whoopsi 😄

handlers/handler.go
```diff
@@ -82,12 +85,20 @@ func InitHandlers(e *echo.Echo) error {
 	e.POST(p+"secret", SetSecret)
 	e.PUT(p+"secret/update", SetSecret)
 
+	// Worker
+	e.GET(p+"worker/secret", GetWorkerRegisterSecret)
+	e.POST(p+"worker/register", RegisterWorker)
```
Skarlso: To follow some practices, this could be just simply `worker` and a `POST` since GET `worker` gets a worker. I don't think you need the `register` part.

helper/stringhelper/stringhelper.go
```diff
@@ -0,0 +1,16 @@
+package stringhelper
+
+import "strings"
+
+// IsContainedInSlice checks if the given string is contained in the slice.
+// If case insensitive is enabled, checks are case insensitive.
+func IsContainedInSlice(s []string, c string, caseIns bool) bool {
```
Skarlso: This is never called with caseIns = false. :)

handlers/pipeline_test.go
```diff
@@ -28,7 +28,7 @@ type mockScheduleService struct {
 	err         error
 }
 
-func (ms *mockScheduleService) SchedulePipeline(p *gaia.Pipeline, args []gaia.Argument) (*gaia.PipelineRun, error) {
+func (ms *mockScheduleService) SchedulePipeline(p *gaia.Pipeline, args []*gaia.Argument) (*gaia.PipelineRun, error) {
```
Skarlso: How come this is now a pointer?

Skarlso:
so this will take a little while... I want to go over it on a high level.. and then start to go over it structurally and then architecturally.... :D But I'll try to hurry things up so it doesn't age too much.

Probably will be fine at first then as we go we work the kinks out.

helper/stringhelper/stringhelper.go
michelvocks: Not yet. I just want to make sure that its there when someone needs it 😃

handlers/pipeline_test.go
michelvocks: I found some inconsistencies when and where we are using pointer or values. I just changed it now to be consistent (usually pointers).

michelvocks:
> so this will take a little while... I want to go over it on a high level.. and then start to go over it structurally and then architecturally.... :D But I'll try to hurry things up so it doesn't age too much.
> 
> Probably will be fine at first then as we go we work the kinks out.

Yep, no problem. A lot of tests are still missing and the documentation.

workers/agent/agent.go
```diff
@@ -0,0 +1,686 @@
...
+
+// InitAgent initiates the agent instance
+func InitAgent(scheduler scheduler.GaiaScheduler, store store.GaiaStore) *Agent {
+	ag := &Agent{
+		scheduler: scheduler,
+		store:     store,
+	}
+
+	// Set path to local certificates
+	ag.certFile = filepath.Join(gaia.Cfg.HomePath, "cert.pem")
```
Skarlso: Shouldn't this be a separate configuration entity since agents can be in a different location then gaia Home?

workers/agent/agent.go
```diff
@@ -0,0 +1,686 @@
...
+	runs, err := a.store.PipelineGetAllRuns()
+	if err != nil {
+		gaia.Cfg.Logger.Error("failed to load pipeline runs from store", "error", err.Error())
+		return
+	}
+
+	// Send all pipeline runs to the remote primary instance
+	for _, run := range runs {
+		// Transform to protobuf struct
+		runPB := &pb.PipelineRun{
```
Skarlso: So, bare with me on this one. What I started to do is instead of creating a struct and assigning the address of it to a variable, I started to do the address of it at the point I want to share it. To clarify, the above code would be like this:

```go
			j := pb.Job{
				UniqueId:    job.ID,
				Title:       job.Title,
				Status:      string(job.Status),
				Description: job.Description,
			}
			runPB.Jobs = append(runPB.Jobs, &j)
```

It's a very subtle change, but makes it clear that at THAT point you are sharing this struct. What say you?

workers/agent/agent.go
```diff
@@ -0,0 +1,686 @@
...
+		}
+	}
+
+	// Set pipeline executable rights
+	return os.Chmod(pipelinePath, 0766)
+}
+
+// updateWork is function that is periodically called and it is used to
+// send new information about a pipeline run to the Gaia master instance.
+func (a *Agent) updateWork() {
```
Skarlso: This and scheduleWork REEALLLY need to be broken down. These functions are way too large man. :)

Skarlso:
Uh, something's not right. The tables are completely out of my window proportions.
Screenshot 2019-06-04 at 22 29 01

Skarlso:
Huh. So I tried creating a pipeline. It said it created it successfully, but it's nowhere to be seen. :D 
So Gaia master can no longer deal with pipelines?

workers/agent/agent.go
michelvocks: Agents use `HomePath` to store their local data which must be different from the primary home directory. I just thought there is no reason to create a new path variable for the agents.

workers/agent/agent.go
michelvocks: I'm wondering if there is some kind of convention? Happy to stick to one solution

handlers/handler.go
michelvocks: Good catch. Thanks!

workers/agent/agent.go
michelvocks: Yeah. I still need to write tests for them anyway so I think I have to refactor them a bit.

helper/filehelper/filehelper_test.go
```diff
@@ -0,0 +1,30 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(tmp)
+	calcSha, err := GetSHA256Sum(filePath)
+	if err != nil {
+		t.Fatal(err)
+	}
+	h := sha256.New()
+	h.Write(sumText)
```
golangcibot: Error return value of `h.Write` is not checked (from `errcheck`)

michelvocks:
@Skarlso I guess you need a bigger screen 😄 
The problem was that the table panel was too big. I renamed the panels and removed "Manage" from them.

workers/agent/agent.go
Skarlso: Not that I know of. I think it's just usage habit. Whatever suits you better I guess. :)

Skarlso:
Alright, looks good now. :)

Skarlso:
@michelvocks how do you register a worker? :D

Skarlso:
Where is the agent code? I thought there will be a gaia systemd daemon somewhere with which Gaia can communicate?

But it looks like Gaia itself should be running on the agent?

Skarlso:
So, I'm trying to register a worker. And I'm getting this:
```
wrong global worker secret provided
```

This means that the secret that is provided in the menu called Worker registration secret: did not equal with that secret. But I can't change it. So.... what's up? :D

Skarlso:
NO idea why this is not working:
`-d '{"secret":"55695ea0-f75e-5b99-b4a3-429c3e3f1b9b","name":"test_worker","tags":["tag1"]}'`

handlers/vault.go
```diff
@@ -37,6 +39,12 @@ func SetSecret(c echo.Context) error {
 		key = s.Key
 		value = s.Value
 	}
+
```
Skarlso: It looks like this has been already put in here however, not:

```go

type addSecret struct {
	Key   string `json:"key"`
	Value string `json:"value"`
}

type updateSecret struct {
	Key   string `json:"key"`
	Value string `json:"newvalue"`
}
```

You don't need to do this. Json accepts multiple values:

```go
type updateSecret struct {
	Key   string `json:"key"`
	Value string `json:"value,newvalue"`
}
```

This way you would only have one struct. :)

Skarlso:
It's getting late. :) Some documentation is necessary dude. :) So I received back the cert, but where do I put it? :D I need to browse the code for that. :) I'm guessing that's why the worker is inactive?

A few suggestions though. I would fail the pipeline if the tag is not found. Saying no matching machines with tag `bla` instead of just running it on master.

Also, display the tags for a worker. Right now I can't see what tags my worker has. :)

With that, good night!

Skarlso:
Okay, I promise I will stop writing on this PR because it's getting huge. :D And I always write all my nuances here. I'll open a Slack channel for debugging / testing this PR and if there is a bug or something I'll comment that. Otherwise this will be a very long PR and hard to follow what's me rambling and an actual bug. :)

handlers/vault.go
michelvocks: You did this. I thought the same when I saw that 😄

michelvocks:
> how do you register a worker? :D

The agent is built into the normal binary. Makes it easier to distribute the binary (only one binary for everything). To register a worker you need to set a few parameters. This is the config I used for testing:
```
hosturl=http://localhost:8080
mode=worker
workername=my-worker
grpchosturl=localhost:8989
workersecret=cfd7b91c-ae5b-5872-9e1f-28592a935ee2
```

> Where is the agent code? I thought there will be a gaia systemd daemon somewhere with which Gaia can communicate?
> 
> But it looks like Gaia itself should be running on the agent?

When you start Gaia, you define with the `mode` parameter in which mode Gaia should start (worker or server).

> It's getting late. :) Some documentation is necessary dude. :) So I received back the cert, but where do I put it? :D I need to browse the code for that. :) I'm guessing that's why the worker is inactive?

Yeah don't worry. Documentation will follow. Just wanted to finish the tests first. :-)
When the worker starts it will automatically look for an existing certificate. If the certificate is not there it will try to register at the primary instance which will attach the mTLS certs in response and place them inside your home folder. So by default, you should do nothing with it since everything should work out of the box. Are there any errors in the logs or something else?

Here is a brief list of steps you need to do that it works:

1. Start Gaia as usual.
2. Optional: Compile the gaia binary and place it in a different folder.
3. Create a config file `.gaia_config` like above (don't forget to replace the secret) in the folder with the binary.
4. Start the worker with the following command: `./gaia -homepath=${PWD}/tmp`

> A few suggestions though. I would fail the pipeline if the tag is not found. Saying no matching machines with tag `bla` instead of just running it on master.

My idea is to add an option which prevents Gaia to run pipelines on the primary instance.

> Also, display the tags for a worker. Right now I can't see what tags my worker has. :)

Good point. Will add that.

handlers/vault.go
Skarlso: OH Gods. What was I thinking? :D :D :D :D

michelvocks:
@Skarlso Okay, I know why the registration process was not working. I changed the registration API endpoint and then it was secured (you needed to be registered). I fixed that by changing the registration path back to `worker/registration`. This is the only way to make this path "insecure" without affecting the other endpoints.

Skarlso:
@michelvocks  Oh I see. That's why it was there. :D Okay.

Skarlso:
@michelvocks BTW. Instead of the horrible, growing endpoint if exclusion thing we have going on, why aren't we using echo's grouping feature? https://echo.labstack.com/guide/routing

michelvocks:
@Skarlso I'm not really getting how the group feature would help us here? As far as I understand the grouping feature, you still have to define every endpoint.

Skarlso:
@michelvocks instead of the if, everything that's in the admin group would require authentication to access. Everything outside is public. No more if.

helper/filehelper/filehelper_test.go
```diff
@@ -0,0 +1,32 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestGetSHA256Sum(t *testing.T) {
+	tmp, _ := ioutil.TempDir("", "TestGetSHA256Sum")
+	sumText := []byte("hello world\n")
+	filePath := filepath.Join(tmp, "test.file")
+	err := ioutil.WriteFile(filePath, sumText, 777)
```
golangcibot: SA9002: file mode '777' evaluates to 01411; did you mean '0777'? (from `staticcheck`)

helper/filehelper/filehelper_test.go
```diff
@@ -0,0 +1,32 @@
...
+	defer os.RemoveAll(tmp)
+	calcSha, err := GetSHA256Sum(filePath)
+	if err != nil {
+		t.Fatal(err)
+	}
+	h := sha256.New()
+	if _, err := h.Write(sumText); err != nil {
+		t.Fatal(err)
+	}
+	if bytes.Compare(h.Sum(nil), calcSha) != 0 {
```
golangcibot: S1004: should use !bytes.Equal(h.Sum(nil), calcSha) instead (from `gosimple`)

workers/server/worker_test.go
```diff
@@ -0,0 +1,319 @@
...
+	md["uniqueid"] = "my-unique-id"
+	return metadata.NewIncomingContext(context.Background(), metadata.New(md))
+}
+
+type mockStreamBinaryServ struct {
+	grpc.ServerStream
+}
+
+func (ms mockStreamBinaryServ) Send(c *pb.FileChunk) error {
+	if bytes.Compare(c.Chunk, []byte("test data")) != 0 {
```
golangcibot: S1004: should use !bytes.Equal(c.Chunk, []byte("test data")) instead (from `gosimple`)

workers/server/worker_test.go
```diff
@@ -0,0 +1,319 @@
...
+	gaia.Cfg = &gaia.Config{}
+	gaia.Cfg.Logger = hclog.New(&hclog.LoggerOptions{
+		Level: hclog.Trace,
+		Name:  "Gaia",
+	})
+	services.MockMemDBService(&mockMemDBService{})
+
+	// Create test pipeline file
+	testPipeline := filepath.Join(tmp, "my-pipeline_golang")
+	if err := ioutil.WriteFile(testPipeline, []byte("test data"), 777); err != nil {
```
golangcibot: SA9002: file mode '777' evaluates to 01411; did you mean '0777'? (from `staticcheck`)

workers/server/worker_test.go
```diff
@@ -0,0 +1,319 @@
...
+
+	// Validate output file
+	logFolderPath := filepath.Join(gaia.Cfg.WorkspacePath, "1", "1", gaia.LogsFolderName)
+	logFilePath := filepath.Join(logFolderPath, gaia.LogsFileName)
+
+	content, err := ioutil.ReadFile(logFilePath)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if bytes.Compare(content, []byte("test log datatest log data")) != 0 {
```
golangcibot: S1004: should use !bytes.Equal(content, []byte("test log datatest log data")) instead (from `gosimple`)

helper/filehelper/filehelper_test.go
Skarlso: I love this one!

michelvocks:
@Skarlso Tests are looking fine now. The test coverage will be decreased a bit but mainly because codecov sucks 😞 I touched a few files where we didn't have any tests and codecov simply counts them now as new code 🙄 

The plan is to finish the documentation part within the next week. Would be great if you could have another look until then. 😄

Skarlso:
@Skarlso sure thing dude. :) It's kind of hard to review 6000 lines of code change though... :D

Skarlso:
Noice!
Screenshot 2019-06-16 at 15 06 46

Skarlso:
@michelvocks Is there a was to delete the default tags? I mean what if my worker isn't supporting the Python environment? :)

Skarlso:
@michelvocks  Also, the pipelines need a way to display their affiliated tags. I can't view what environment my pipeline needs once I set a tag.

michelvocks:
> @Skarlso sure thing dude. :) It's kind of hard to review 7700 lines of code change though... :D

I don't expect you to review all the lines of code. 😄 
It already helps a lot if you just have a general look and may provide feedback like above. Super helpful 🤗 

> @michelvocks Is there a was to delete the default tags? I mean what if my worker isn't supporting the Python environment? :)

I've added the possibility to provide "negative tags". So if you provide a tag like "-golang" it will remove automatically the "golang" tag (and the "-golang" tag). Not the best solution in terms of user experience but should work for this use-case.  😄 

> @michelvocks Also, the pipelines need a way to display their affiliated tags. I can't view what environment my pipeline needs once I set a tag. I thought that maybe we'd need the ability to edit tags, but I reconsidered. In case an environment need changes for a pipeline we'd just delete it and re-create it. But we still need an ability to see them tags. :)

Whops, yeah. I added the tags to the overview page.

Skarlso:
@michelvocks awesome, I'm okay with the negative tags. :D so do you tag a pipeline automatically when it's created with the appropriate environment tag? Like a Go pipeline gets go a Python pipeline gets Python etc?

michelvocks:
> @michelvocks awesome, I'm okay with the negative tags. :D so do you tag a pipeline automatically when it's created with the appropriate environment tag? Like a Go pipeline gets go a Python pipeline gets Python etc?

Yep: https://github.com/gaia-pipeline/gaia/pull/166/files#diff-f88a32f252fb4994c4b7249afcc79b6cR57

michelvocks:
@Skarlso Docs PR: https://github.com/gaia-pipeline/docs/pull/11

Still WIP. Need to update the RFC.

michelvocks:
@Skarlso Docs PR should be fine now. In my opinion, this is now ready for merge.

Let me know what you think. 🤗

workers/agent/agent.go
```diff
@@ -0,0 +1,713 @@
...
+}
+
+// setupConnectionInfo setups the connection info object by parsing existing
+// mTLS certificates or registering the worker at the Gaia primary instance
+// and receiving new mTLS certs.
+func (a *Agent) setupConnectionInfo() (credentials.TransportCredentials, error) {
+	// Evaluate and summarize all worker tags
+	var tags []string
+	if gaia.Cfg.WorkerTags != "" {
+		trimmedTags := strings.ReplaceAll(gaia.Cfg.WorkerTags, " ", "")
```
Skarlso: @michelvocks hm, this means that we aren't allowing for tags with spaces in them. That is okay, I guess, right?

michelvocks: Correct. Tags are single words which can be combined with dashes (for example). From my experience, that is the expected behavior.

Skarlso:
@michelvocks Well. Functionality wise this looks okay. I tested it and it worked. :D But man, the code needs refactoring. :DDD Create a bunch of follow up tickets so we can track the most common ones? Like the functions are too long and such? Otherwise, go ahead. :)

michelvocks:
> @michelvocks Well. Functionality wise this looks okay. I tested it and it worked. :D But man, the code needs refactoring. :DDD Create a bunch of follow up tickets so we can track the most common ones? Like the functions are too long and such? Otherwise, go ahead. :)

Is it only the length of some functions? In my opinion, it is quite common and normal to have a bit longer functions. It makes it often more readable because you can simply follow one block of code instead of switching between several functions. I also know other big projects (Kubernetes, Vault, Terraform etc.) that bigger functions can be beneficial.

michelvocks:
ping @Skarlso

Skarlso:
I disagree with that. That is why cyclomatic complexity is a thing. And that is why lingers are saying that your functions is too long. It's harder to track down a bug in a big ass function than in a small function where it's basically clear at one quick look what it does. If I have scroll down several pages and still be in the same function I'll loose track of what the beginning of the function was even trying to do in the first place. 😂 Also if a function is huge it's more likely that it does not adhere to SRP. The single responsibility principle. It is likely however that its trying to do everything / too much of its own.

I approve but we have to refractor these into easily tracakble understandable fragments I think. 
😊

Also, sorry for the late response, I'm on vacation in Austria. 😁

PR #168, opened 2019-02-03
[WIP] Add permissions groups store, handlers and login functionality

### LINKED ISSUE #164: [Permissions] Add the ability to define permission groups
[Permissions] Add the ability to define permission groups

There is now a basic permissions system in the form of user roles. We should add the ability to create permission groups where the roles can be configured. Users can then be added or removed from groups, effectively granting them the extra permissions of the groups they are part of.

speza:
A WIP. Will eventually resolve #164.


Skarlso:
@speza Hey man. :) Do you have anything more to add to this by any chance? :)

Sorry, there is now a lot of merge conflict on this. :)

speza:
Hey @Skarlso! Good question! Apologies, life has gotten in the way of doing anything open source at the moment. 

I guess this comes down to whether we need this enhanced functionality at this stage?

I'd love to get back into it at some point. So I could attempt to fix them sometime or rework the whole change. Its been a while since I delved into here though.

Skarlso:
I think it's a good to have thing if we want to take things seriously. :) So if you need any help just let me know. Also, I'm very sorry, but I'm about to merge a large change to the singleton service thing. :D

speza:
Agreed @Skarlso :) Thank you, will do.
And ha np. I will be honest, I don't favour the singleton approach in general :stuck_out_tongue:

Skarlso:
@speza Oh definitely, I am not so happy about it either. But at that time I thought it's cool. Time passed, and I would like to think that I learned a lot since. :) So now I see better approaches which I would like to apply with providers.

But it's a lot of work and a massive refactor since these services are so deeply engrained. I'll get there in time... This is a first step. The scheduler was the smallest and resulted in a TON of change... :D

speza:
Yeah no worries @Skarlso! I also didn't mention anything before because I thought it was ok too :)

I'm a big fan of proper dependency injection wherever I write code these days, primarily Java still (ew) but been a lot more Go too recently. I favour building the dependency higher up in the application and passing it all the way through where required. Still been meaning to try out something like https://github.com/google/wire - but in general I follow the pattern highlighted in their tutorial (just without the wire generated code).

speza:
Closing. Working on a better solution :)

PR #172, opened 2019-02-15
Fix git SSH clone in docker container

### LINKED ISSUE #149: Problem using git@ uri: "SSH agent requested but SSH_AUTH_SOCK not-specified"
Hi, 
im getting the above error when trying to add a pipeline with "git@...". Im running the `gaiapipeline/gaia:latest` container, i configured `/root/.ssh/id_rsa` and pub key, so if i exec the git-clone inside of the container, it works fine. Any hints? Google mostly points me to something with windows and putty, which isnt the case for me (running on k8s cluster on gce). The `PipelineGitLSRemote` seems not to be very verbose, so i guess the error is thrown further down, is there anything i can do to get more insight?

image

michelvocks:
Fixes #149 

@Skarlso Feel free to merge if it's okay from your side. 🤗

### DIFF
--- a/workers/pipeline/git.go
+++ b/workers/pipeline/git.go
@@ -59,7 +59,7 @@ func GitLSRemote(repo *gaia.GitRepo) error {
 	s, err := cl.NewUploadPackSession(ep, auth)
 	if err != nil {
 		if strings.Contains(err.Error(), "knownhosts: key is unknown") {
-			gaia.Cfg.Logger.Error("Warning: Unknown host key.", "error", err.Error(), "URL", repo.URL)
+			gaia.Cfg.Logger.Warn("Warning: Unknown host key.", "error", err.Error(), "URL", repo.URL)
 			auth, err := getAuthInfo(repo, gossh.InsecureIgnoreHostKey())
 			if err != nil {
 				return err
@@ -120,7 +120,7 @@ func UpdateRepository(pipe *gaia.Pipeline) error {
 	err = tree.Pull(o)
 	if err != nil {
 		if strings.Contains(err.Error(), "knownhosts: key is unknown") {
-			gaia.Cfg.Logger.Error("Warning: Unknown host key.", "error", err.Error(), "host", "URL", pipe.Repo.URL)
+			gaia.Cfg.Logger.Warn("Warning: Unknown host key.", "error", err.Error(), "host", "URL", pipe.Repo.URL)
 			auth, err = getAuthInfo(&pipe.Repo, gossh.InsecureIgnoreHostKey())
 			if err != nil {
 				return err
@@ -166,7 +166,7 @@ func gitCloneRepo(repo *gaia.GitRepo) error {
 	_, err = git.PlainClone(repo.LocalDest, false, o)
 	if err != nil {
 		if strings.Contains(err.Error(), "knownhosts: key is unknown") {
-			gaia.Cfg.Logger.Error("Warning: Unknown host key.", "error", err.Error(), "URL", repo.URL)
+			gaia.Cfg.Logger.Warn("Warning: Unknown host key.", "error", err.Error(), "URL", repo.URL)
 			auth, err = getAuthInfo(repo, gossh.InsecureIgnoreHostKey())
 			if err != nil {
 				return err
--- a/workers/pipeline/git.go
+++ b/workers/pipeline/git.go
@@ -59,9 +59,11 @@ func GitLSRemote(repo *gaia.GitRepo) error {
 	s, err := cl.NewUploadPackSession(ep, auth)
 	if err != nil {
 		if strings.Contains(err.Error(), "knownhosts: key is unknown") {
-			gaia.Cfg.Logger.Error("Warning: Unknown host key.")
-			auth, _ := getAuthInfo(repo, gossh.InsecureIgnoreHostKey())
-			err = nil
+			gaia.Cfg.Logger.Error("Warning: Unknown host key.", "error", err.Error(), "URL", repo.URL)
+			auth, err := getAuthInfo(repo, gossh.InsecureIgnoreHostKey())
+			if err != nil {
+				return err
+			}
 			s, err = cl.NewUploadPackSession(ep, auth)
 			if err != nil {
 				return err
@@ -118,9 +120,11 @@ func UpdateRepository(pipe *gaia.Pipeline) error {
 	err = tree.Pull(o)
 	if err != nil {
 		if strings.Contains(err.Error(), "knownhosts: key is unknown") {
-			gaia.Cfg.Logger.Error("Warning: Unknown host key.")
-			err = nil
-			auth, _ = getAuthInfo(&pipe.Repo, gossh.InsecureIgnoreHostKey())
+			gaia.Cfg.Logger.Error("Warning: Unknown host key.", "error", err.Error(), "host", "URL", pipe.Repo.URL)
+			auth, err = getAuthInfo(&pipe.Repo, gossh.InsecureIgnoreHostKey())
+			if err != nil {
+				return err
+			}
 			o.Auth = auth
 			err = tree.Pull(o)
 			if err != nil {
@@ -162,9 +166,11 @@ func gitCloneRepo(repo *gaia.GitRepo) error {
 	_, err = git.PlainClone(repo.LocalDest, false, o)
 	if err != nil {
 		if strings.Contains(err.Error(), "knownhosts: key is unknown") {
-			gaia.Cfg.Logger.Error("Warning: Unknown host key.")
-			err = nil
-			auth, _ = getAuthInfo(repo, gossh.InsecureIgnoreHostKey())
+			gaia.Cfg.Logger.Error("Warning: Unknown host key.", "error", err.Error(), "URL", repo.URL)
+			auth, err = getAuthInfo(repo, gossh.InsecureIgnoreHostKey())
+			if err != nil {
+				return err
+			}
 			o.Auth = auth
 			// Clone repo again with no host key verification.
 			_, err = git.PlainClone(repo.LocalDest, false, o)
--- a/workers/pipeline/git.go
+++ b/workers/pipeline/git.go
@@ -5,7 +5,6 @@ import (
 	"crypto/rand"
 	"encoding/base64"
 	"errors"
-	"log"
 	gohttp "net/http"
 	"path"
 	"regexp"
@@ -15,6 +14,7 @@ import (
 	"github.com/gaia-pipeline/gaia"
 	"github.com/gaia-pipeline/gaia/services"
 	"github.com/google/go-github/github"
+	gossh "golang.org/x/crypto/ssh"
 	"golang.org/x/oauth2"
 	git "gopkg.in/src-d/go-git.v4"
 	"gopkg.in/src-d/go-git.v4/plumbing"
@@ -44,7 +44,7 @@ func GitLSRemote(repo *gaia.GitRepo) error {
 	}
 
 	// Attach credentials if provided
-	auth, err := getAuthInfo(repo)
+	auth, err := getAuthInfo(repo, nil)
 	if err != nil {
 		return err
 	}
@@ -60,7 +60,12 @@ func GitLSRemote(repo *gaia.GitRepo) error {
 	if err != nil {
 		if strings.Contains(err.Error(), "knownhosts: key is unknown") {
 			gaia.Cfg.Logger.Error("Warning: Unknown host key.")
+			auth, _ := getAuthInfo(repo, gossh.InsecureIgnoreHostKey())
 			err = nil
+			s, err = cl.NewUploadPackSession(ep, auth)
+			if err != nil {
+				return err
+			}
 		} else {
 			return err
 		}
@@ -89,7 +94,6 @@ func GitLSRemote(repo *gaia.GitRepo) error {
 // it by pulling in new code if it's available.
 func UpdateRepository(pipe *gaia.Pipeline) error {
 	r, err := git.PlainOpen(pipe.Repo.LocalDest)
-	log.Println(pipe.Repo.LocalDest)
 	if err != nil {
 		// We don't stop gaia working because of an automated update failed.
 		// So we just move on.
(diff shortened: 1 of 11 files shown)


workers/pipeline/git.go
```diff
@@ -269,6 +275,9 @@ func getAuthInfo(repo *gaia.GitRepo) (transport.AuthMethod, error) {
 		if err != nil {
 			return nil, err
 		}
+
+		// Ignore host key check for the user experience
+		auth.(*ssh.PublicKeys).HostKeyCallback = ssh2.InsecureIgnoreHostKey()
```
Skarlso: @michelvocks  I would advice against doing this. If it's a public repo, you should use https... other wise, you should add the host to hostkeys. I didn't want to see something like this in production, don't you think? :)

Skarlso: We should validate known_hosts file or use FixedHostKey

michelvocks: @Skarlso This relates to SSH git clone. SSH doesn't use HTTPS since it has its own protocol. Usually, the SSH agent checks if this host is known to prevent man-in-the-middle attacks.

I personally think we can go without this because it makes the setup inside a docker container way more complicated. Additionally, this also the default behavior of Jenkins: https://support.cloudbees.com/hc/en-us/articles/115000073552-Host-Key-Verification-for-SSH-Agents

Skarlso: What I'm saying is if the host is unknown or insecure then rather then using git@ the user should use https or add the key to known_hosts. Also jenkins is a festering pile of security issues. Every week there is a new issue. 😂 I wouldn't use them as an example of good practice regarding security, don't you think?

leberknecht: Ahoi! What would be a clean solution then? I guess we would require user interaction at this point (so more or less "copy the console behavior"), asking for permission, add the host. Maybe it would be a good idea to configure the container to use the known hosts not under `~/.ssh/known_hosts` but on a path of the application, which the user might want to mount anyways (e.g. the user might want to mount a volume on the pipeline-config directory to keep the state).

leberknecht: Hm, plus would need error handling in case the host ip changes.. meh.

Skarlso: I would suggest looking into hostkey verification. The ssh package has this: [ssh.ParseKnownHosts](https://godoc.org/golang.org/x/crypto/ssh#ParseKnownHosts) which might provide some insight. You could also parse out host keys and then use FixedHostKey like this:

```
hostKey := getHostKey(url)
...
HostKeyCallback: ssh.FixedHostKey(hostKey),
```

There are several posts around that talk about SSH seucirty and how checking host source is important so I won't go into that here.

Some links:

https://utcc.utoronto.ca/~cks/space/blog/programming/GoSSHHostKeyCheckingNotes
https://bridge.grumpy-troll.org/2017/04/golang-ssh-security/

michelvocks: To be honest, I don't think is really a security issue:

```
I’ve yet to find a project which actually bothered implementing hostkey verification in SSH client code. None.
```

This actually confirms my acceptance. We should document that for sure.

Skarlso: But the fact that they don't isn't a good thing. :D And we good lead by good example. :))

But of course it's up to you man. :) I realise that this isn't trivial.

Skarlso: @michelvocks @leberknecht  Hey guys. I wrote this thing: https://gist.github.com/Skarlso/34321a230cf0245018288686c9e70b2d

If the entry is hostname based this will work even if the IP changes underneath and there is no new entry for the new IP. This will also fail if there is no entry in the known_hosts file.

This can be easily modified to automatically add one for example... But I would prefer the user managing the host file instead of gaia doing it automatically.

What say you?

leberknecht: I think i would be fine with a speaking error message and a small "how to configure the known hosts for containers / k8s" hint in the documentation :)

Even though it could feel a bit "unusual" to the user (as in "its normal to configure the ssh key for cloning somewhere, as its in the nature of things, but i never needed to configure the known-hosts on any other software before"). But i totally agree that copying "bad behavior from other software" is semi cool :) 

What would be the downside of doing the automatic handling? I guess that we presumably take user agreement to modify the known hosts as given?

Skarlso: You never had to because at some point in time you had a question appear when you first cloned something that asked if you would like to add git to your host file. Try removing git from your known hosts and clone something. It will ask you to add it.

You indirectly configure it. 🙂

michelvocks: The problem, which I have personally experienced, is when you have your Git server behind a load balancer and when you have immutable infrastructure. 

Gitlab, for example, allows you to run it HA. Which basically means you have multiple server and any server can randomly handle the traffic. If you combine that with immutable infrastructure (e.g. when you upgrade your server or your software which is running on your server) you basically screwed because each time you replace your server it will have a different host key. 

In my opinion, SSH host key validation is important and right, but it is simply nowadays with the Cloud not applicable anymore. 

@Skarlso  @leberknecht What do you think about checking the know_hosts file and if there is something suspicious we will prompt a message in the UI. This is still quite user-friendly in my opinion.

Skarlso: A fair point. :) Okay, I'll agree to that. Implementing a hook which updates the known_hosts file whenever there is a new entry for a given host might be a bit out of the scope I guess. :D 👍

Skarlso: BTW in an HA environment, for a given server family you just copy the SSH certificate to the other servers.
These two:
```
/etc/ssh/ssh_host_rsa_key
/etc/ssh/ssh_host_rsa_key.pub
```

This way once a load balancer is added to known_hosts everything underneath it will forever work. The load balancer, when it's rotating in a new instance will automatically copy these keys to that instance. Once that's done, that server will also work without the need to change the known_hosts file.

Skarlso: Hang on. I think I have a suitable fix for this. Gonna push up something soon.

Skarlso:
@michelvocks @leberknecht Alright. What do you guys think? We are checking the host key, but if there is a problem with it, we move on but Log a Warning error message.

In order for this to be visible on the UI though, we would need to return the error, which stops the flow right there. Which is not what we want if I understand correctly.

Skarlso:
I'll add a test soon... daughter woke up.. :)

michelvocks:
This is perfect in my opinion. Thanks @Skarlso 🤗 

I gave it a try in a docker container but it fails with the following panic:
```
2019-02-20T14:16:45.766Z [ERROR] Gaia: Warning: Unknown host key.
{"time":"2019-02-20T14:16:45.7667359Z","level":"-","prefix":"echo","file":"recover.go","line":"73","message":"[PANIC RECOVER] runtime error: invalid memory address or nil pointer dereference goroutine 41 [running]:\ngithub.com/labstack/echo/middleware.RecoverWithConfig.func1.1.1(0x21e47c8, 0x1000, 0xc000500000, 0x229d240, 0xc0001ce000)\n\t/Users/michelvocks/go/pkg/mod/github.com/labstack/echo@v3.3.10+incompatible/middleware/recover.go:71 +0xf8\npanic(0xc1eaa0, 0x285d5d0)\n\t/usr/local/Cellar/go/1.11.4/libexec/src/runtime/panic.go:513 +0x1b9\ngopkg.in/src-d/go-git.v4/plumbing/transport/internal/common.(*session).finish(0x0, 0x0, 0x0)\n\t/Users/michelvocks/go/pkg/mod/gopkg.in/src-d/go-git.v4@v4.5.0/plumbing/transport/internal/common/common.go:334 +0x26\ngopkg.in/src-d/go-git.v4/plumbing/transport/internal/common.(*session).Close(0x0, 0x0, 0x0)\n\t/Users/michelvocks/go/pkg/mod/gopkg.in/src-d/go-git.v4@v4.5.0/plumbing/transport/internal/common/common.go:352 +0x37\npanic(0xc1eaa0, 0x285d5d0)\n\t/usr/local/Cellar/go/1.11.4/libexec/src/runtime/panic.go:513 +0x1b9\ngopkg.in/src-d/go-git.v4/plumbing/transport/internal/common.(*session).AdvertisedReferences(0x0, 0x228fd20, 0x0, 0x0)\n\t/Users/michelvocks/go/pkg/mod/gopkg.in/src-d/go-git.v4@v4.5.0/plumbing/transport/internal/common/common.go:167 +0x26\ngithub.com/gaia-pipeline/gaia/workers/pipeline.GitLSRemote(0xc0002a6aa0, 0x0, 0x0)\n\t/Users/michelvocks/go/src/github.com/gaia-pipeline/gaia/workers/pipeline/git.go:71 +0x20b\ngithub.com/gaia-pipeline/gaia/handlers.PipelineGitLSRemote(0x229d240, 0xc0001ce000, 0xc0004fc498, 0x0)\n\t/Users/michelvocks/go/src/github.com/gaia-pipeline/gaia/handlers/pipeline.go:27 +0x81\ngithub.com/labstack/echo.(*Echo).Add.func1(0x229d240, 0xc0001ce000, 0xc0001acc60, 0xc0000b2f60)\n\t/Users/michelvocks/go/pkg/mod/github.com/labstack/echo@v3.3.10+incompatible/echo.go:490 +0x8a\ngithub.com/gaia-pipeline/gaia/handlers.AuthMiddleware.func1.1(0x229d240, 0xc0001ce000, 0xc00000a140, 0xc3f180)\n\t/Users/michelvocks/go/src/github.com/gaia-pipeline/gaia/handlers/auth.go:54 +0x408\ngithub.com/labstack/echo/middleware.BodyLimitWithConfig.func1.1(0x229d240, 0xc0001ce000, 0x0, 0x0)\n\t/Users/michelvocks/go/pkg/mod/github.com/labstack/echo@v3.3.10+incompatible/middleware/body_limit.go:87 +0x187\ngithub.com/labstack/echo/middleware.RecoverWithConfig.func1.1(0x229d240, 0xc0001ce000, 0x0, 0x0)\n\t/Users/michelvocks/go/pkg/mod/github.com/labstack/echo@v3.3.10+incompatible/middleware/recover.go:78 +0xd8\ngithub.com/labstack/echo.(*Echo).ServeHTTP(0xc00040c000, 0x2290280, 0xc00021fa40, 0xc0001dfc00)\n\t/Users/michelvocks/go/pkg/mod/github.com/labstack/echo@v3.3.10+incompatible/echo.go:593 +0x220\nnet/http.serverHandler.ServeHTTP(0xc000098680, 0x2290280, 0xc00021fa40, 0xc0001dfc00)\n\t/usr/local/Cellar/go/1.11.4/libexec/src/net/http/server.go:2741 +0xab\nnet/http.(*conn).serve(0xc0002a60a0, 0x22910c0, 0xc0002ba040)\n\t/usr/local/Cellar/go/1.11.4/libexec/src/net/http/server.go:1847 +0x646\ncreated by net/http.(*Server).Serve\n\t/usr/local/Cellar/go/1.11.4/libexec/src/net/http/server.go:2851 +0x2f5\n\ngoroutine 1 [IO wait, 2 minutes]:\ninternal/poll.runtime_pollWait(0x7f950630df00, 0x72, 0x0)\n\t/usr/local/Cellar/go/1.11.4/libexec/src/runtime/netpoll.go:173 +0x66\ninternal/poll.(*pollDesc).wait(0xc00020e718, 0x72, 0xc0001e0000, 0x0, 0x0)\n\t/usr/local/Cellar/go/1.11.4/libexec/src/internal/poll/fd_poll_runtime.go:85 +0x9a\ninternal/poll.(*pollDesc).waitRead(0xc00020e718, 0xffffffffffffff00, 0x0, 0x0)\n\t/usr/local/Cellar/go/1.11.4/libexec/src/internal/poll/fd_poll_runtime.go:90 +0x3d\ninternal/poll.(*FD).Accept(0xc00020e700, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)\n\t/usr/local/Cellar/go/1.11.4/libexec/src/internal/poll/fd_unix.go:384 +0x1a0\nnet.(*netFD).accept(0xc00020e700, 0x203000, 0x0, 0x0)\n\t/usr/local/Cellar/go/1.11.4/libexec/src/net/fd_unix.go:238 +0x42\nnet.(*TCPListener).accept(0xc0001ea2d0, 0xc000044f00, 0x40b77f, 0xc0002f80a0)\n\t/usr/local/Cellar/go/1.11.4/libexec/src/net/tcpsock_posix.go:139 +0x2e\nnet.(*TCPListener).AcceptTCP(0xc0001ea2d0, 0xfba3a5d3, 0x89ad5208cca40317, 0x4491b8)\n\t/usr/local/Cellar/go/1.11.4/libexec/src/net/tcpsock.go:247 +0x47\ngithub.com/labstack/echo.tcpKeepAliveListener.Accept(0xc0001ea2d0, 0xc0001f7b50, 0x47b846, 0x5c6d60d5, 0x433fce)\n\t/Users/michelvocks/go/pkg/mod/github.com/labstack/echo@v3.3.10+in\n"}
```

I think the problem is that `s, err := cl.NewUploadPackSession(ep, auth)` returns an error and `s` will be `nil`.

Skarlso:
@MichelVocks intersting

Skarlso:
Ups... So... Interesting. Thanks! I'll check it out and fix it. 😊

Skarlso:
Ah, I see. I thought it would just ignore missing files. But it doesn't. :) Fixing it. :)

Skarlso:
Oh I see what you mean...

Skarlso:
@michelvocks It was worse than that. :D I was checking in the wrong place. Basically, since the operation failed with an invalid host key.. I have to retry said operation again with an Auth that is using an Ignore.

It looks a bit more verbose this way though, check it out. I'll push in a bit.

michelvocks:
It works now @Skarlso. I did some minor changes to make the logging a bit more verbose and added an extra small check.

Skarlso:
@MichelVocks looks good. I'm unsure yet how I'll test this as it finds my local known_hosts file. I'll try to mock the hostname from where it will clone. Maybe that'll work.

Skarlso:
Note: Working on writing / fixing test for GitLS

Skarlso:
@michelvocks first passing test!!! :)))))))

Skarlso:
@michelvocks alright. Good as it gets. decrease by 0.2%. :D

Skarlso:
Yay, I managed to increase coverage. :)

michelvocks:
Thanks @Skarlso. Awesome job! 🤗

PR #174, opened 2019-02-23
Toggle Polling from the UI

Skarlso:


Skarlso:
@michelvocks I might have to pick your brain on how to go about with a Toggle button. I want to get its state and set it to green or something based on the state of the config. And then toggling that will change the config value.

michelvocks:
> @michelvocks I might have to pick your brain on how to go about with a Toggle button. I want to get its state and set it to green or something based on the state of the config. And then toggling that will change the config value.

You can use [vue-bulma-switch](https://github.com/vue-bulma/switch) for this. It's pretty old and not really updated but basically all our current components are based on vue-bulma so I think it makes sense.

Let me know if you need something else.

workers/pipeline/ticker.go
```diff
@@ -22,6 +23,47 @@ const (
 	tickerIntervalSeconds = 5
 )
 
+var pollerDone = make(chan struct{}, 0)
+var isPollerRunning bool
+
+// StopPoller sends a done signal to the polling timer if it's running.
+func StopPoller() error {
+	if isPollerRunning {
+		pollerDone <- struct{}{}
```
Skarlso: @skarlso you need to set isRunning to false here.

Skarlso: Sorry, this was a reminder for me. :D

Skarlso:
> > @michelvocks I might have to pick your brain on how to go about with a Toggle button. I want to get its state and set it to green or something based on the state of the config. And then toggling that will change the config value.
> 
> You can use [vue-bulma-switch](https://github.com/vue-bulma/switch) for this. It's pretty old and not really updated but basically all our current components are based on vue-bulma so I think it makes sense.
> 
> Let me know if you need something else.

Cool, thanks!

Skarlso:
Uh, we will have to eventually take this apart into smaller files or something. Navigating this view is becoming difficult. :) I don't know how vue works but that should be possible. :)

Skarlso:
Oh! I see how permissions did it. I'll do that same. :) Extract settings into there. Nice.

Skarlso:
@michelvocks This looks like shit right now. :D But actually works. :) the toggle and all work together stopping / starting the poller. :) And the button is set to on and off on mounting so it shows what the value currently is. Which is nice. :D
screenshot 2019-02-24 at 21 40 15
screenshot 2019-02-24 at 21 43 04

I also tried turning it on and off a hundred times or so quickly to see if I can break the underlying channel logic. But seems to work okay. :)

I'll try and come up with a design for this. And write some handler tests. :)

michelvocks:
@Skarlso Cool. Looking forward to trying it out. This is also a solid foundation for additional settings 👍

Skarlso:
@michelvocks yeah I'm trying to work out a better way of doing this 🤔 Maybe save all the possible settings in a map pointer with string keys or something. I'll do some more thinking on it. It would be cool to have something common that all the settingon-off methods could use.

You still would have get set methods for individual settings but they would use something common. But maybe it's not worth it. Don't know until I have a go at it. :) I don't want to use reflection either.

michelvocks:
@Skarlso I think you can just use a struct like we normally do: https://github.com/gaia-pipeline/gaia/blob/master/frontend/client/views/pipeline/create.vue#L336

Skarlso:
Yes. But the settings are of different types. And I sort of thought for something like: setSetting(name, key, string) bamm, done. :D But that's not going to work. Unless we use something like, flag.Func or something like that which takes a function and would handle conversion. Not sure yet.

michelvocks:
I think it's easier to send all settings in a struct and check in the backend what has changed.

Skarlso:
I was thinking about that too to just always send everything. But I don't like that. :) I would like to keep the Json size to a bare minimum. If we always send / get everything that will increase the amount of bandwidth Gaia is using.

Skarlso:
Eh, I don't know. Maybe it's okay for now. I don't see how we'd change some of the other config settings like port of server address or working directory without the need to restart Gaia anyways.

Skarlso:
@michelvocks found a nicer button ;)
screenshot 2019-02-26 at 21 47 37
screenshot 2019-02-26 at 21 47 32

Easier to change programatically too.

Skarlso:
@michelvocks Dude. I made it work!! :))))

screenshot 2019-02-26 at 22 38 13

michelvocks:
Nice! 🤗

Skarlso:
@michelvocks Read for you man. :) Increases coverage \o/

frontend/client/views/settings/index.vue
```diff
@@ -103,6 +103,9 @@
             </div>
           </div>
         </tab-pane>
+        <tab-pane label="Manage Settings" icon="fa fa-wrench">
```
Skarlso: Ah I probably need a different icon... :)

michelvocks: Wrench looks good in my opinion. What were you thinking of?

frontend/client/views/settings/settings/manage-settings.vue
```diff
@@ -0,0 +1,154 @@
+<template>
```
michelvocks: Thanks for using templates. I should have done that way earlier 👍

frontend/client/views/settings/settings/manage-settings.vue
```diff
@@ -0,0 +1,154 @@
...
+
+  Vue.use(VueGoodTable)
+  Vue.use(VueTippy)
+
+  export default {
+    name: 'manage-settings',
+    components: {Tabs, TabPane, ToggleButton},
+    data () {
+      return {
+        // search: '',
```
michelvocks: Delete if it's not needed 😄

michelvocks:
@Skarlso One additional thing we should think about:

What happens when you restart Gaia or it crashes? The change you made in the UI is basically lost. 
This might be frustrating for users. We could save the change in the boltdb but then we have to decide which change takes precedence in case the user set an environment variable or used a configuration file.

Skarlso:
@michelvocks Our flag implementation now supports environment files. We could write this out in that file which does take precedence.

frontend/client/views/settings/index.vue
Skarlso: ¯\_(ツ)_/¯ Don't know. It was just interesting that there is another one already with the wrench.

Skarlso:
Uh, the file approach sucks because I can't really overwrite. I can search and append, or fseek. But it isn't particularly pretty. :/ 

I will have to resort to the DB.

Skarlso:
Oh yeah so the question is, which one would be used first.... I'm going with the bbolt setting being of higher importance.

frontend/client/views/settings/index.vue
michelvocks: Ah I see. Manage Pipelines has it already. Maybe you can try `fa-cog`? https://fontawesome.com/icons?d=gallery&q=settings

handlers/settings.go
```diff
@@ -0,0 +1,62 @@
...
+
+// SettingsPollOn turn on polling functionality.
+func SettingsPollOn(c echo.Context) error {
+	storeService, _ := services.StorageService()
+	gaia.Cfg.Poll = true
+	err := pipeline.StartPoller()
+	if err != nil {
+		c.String(http.StatusBadRequest, err.Error())
+	}
+	configStore := &gaia.StoreConfig{}
```
michelvocks: We should load it in case we have future settings and we don't want to overwrite them here.

handlers/settings.go
```diff
@@ -0,0 +1,62 @@
...
+
+// SettingsPollOff turn off polling functionality.
+func SettingsPollOff(c echo.Context) error {
+	storeService, _ := services.StorageService()
+	gaia.Cfg.Poll = false
+	err := pipeline.StopPoller()
+	if err != nil {
+		c.String(http.StatusBadRequest, err.Error())
+	}
+	configStore := &gaia.StoreConfig{}
```
michelvocks: We should load it in case we have future settings and we don't want to overwrite them here.

store/pipeline.go
```diff
@@ -323,3 +323,46 @@ func (s *BoltStore) PipelineDelete(id int) error {
 		return b.Delete(itob(id))
 	})
 }
+
+// SettingsPut puts settings into the store.
+func (s *BoltStore) SettingsPut(c *gaia.StoreConfig) error {
```
michelvocks: We should put that into it's own file e.g. store/settings.go

Skarlso: Absolutely. I'm just playing around here seeing how it would work.

handlers/settings.go
```diff
@@ -0,0 +1,79 @@
...
+	return c.String(http.StatusOK, "Polling is turned off.")
+}
+
+type pollStatus struct {
+	Status bool
+}
+
+// SettingsPollGet get status of polling functionality.
+func SettingsPollGet(c echo.Context) error {
+	storeService, _ := services.StorageService()
```
Skarlso: got to check that error here

handlers/settings.go
```diff
@@ -0,0 +1,79 @@
...
+	err = storeService.SettingsPut(configStore)
+	if err != nil {
+		c.String(http.StatusBadRequest, err.Error())
+	}
+	return c.String(http.StatusOK, "Polling is turned on.")
+}
+
+// SettingsPollOff turn off polling functionality.
+func SettingsPollOff(c echo.Context) error {
+	storeService, _ := services.StorageService()
```
Skarlso: and here

Skarlso:
@michelvocks alright. this is ready for review again. I tested it manually too. Turned it on, quit Gaia, started it up again, and setting was still on. :)

PR #175, opened 2019-02-27
Remove the channel Iterator method from ActiavePipelines

### LINKED ISSUE #163: Remove the Iter function from ActivePipelines
It doesn't really provide any sane functionality, since it's linear but adds traps and confusion since the channel cannot be broken out of. Which means if a pipeline is not found, we need to check names instead of nil. Or the fact that it has to always go through all of them all the time. This wouldn't be true if we had an additional way to signal a cancel, but that just adds complexity without real benefit.

Just get rid of it.

Skarlso:
Closes #163

### DIFF
--- a/workers/pipeline/pipeline.go
+++ b/workers/pipeline/pipeline.go
@@ -133,7 +133,7 @@ func (ap *ActivePipelines) Remove(index int) {
 
 // GetByName looks up the pipeline by the given name.
 func (ap *ActivePipelines) GetByName(n string) *gaia.Pipeline {
-	for _, pipeline := range ap.Iter() {
+	for _, pipeline := range ap.GetAll() {
 		if pipeline.Name == n {
 			return &pipeline
 		}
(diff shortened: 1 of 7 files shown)


workers/pipeline/pipeline.go
```diff
@@ -174,62 +169,43 @@ func (ap *ActivePipelines) Replace(p gaia.Pipeline) bool {
...
-	go func() {
-		ap.RLock()
-		defer ap.RUnlock()
-		for _, pipeline := range ap.Pipelines {
-			c <- pipeline
-		}
-		close(c)
-	}()
-
+func (ap *ActivePipelines) Iter() []gaia.Pipeline {
```
michelvocks: `Iter` sounds like you get an iterator back. Maybe `GetAll` or `ToSlice` is better?

Skarlso: @michelvocks I didn't wanted to change the whole thing at once... I thought I'll keep the name for now, to see if anything breaks.

But something I didn't realise is that I'm already giving back a list. Why am I going over it again in some situations. :D Made no sense. thanks for spotting that.

I'll rename it. :)

Issue #160, opened 2019-01-29
Add go linter to circleci


michelvocks:
We are not enforcing the `go linter` (https://github.com/golang/lint) currently during circleci build. We should do that.

PR #176, opened 2019-03-03
Adding linter to circle CI.

Skarlso:


Skarlso:
@michelvocks Hi. :) The changes in here are because of the new linter. :D I had to rename the GRPC struct and added a comment. The renaming affected some files though. I checked, and they shouldn't be used by anything else in Gaia like the SDK for example.

Skarlso:
Closes #160

PR #183, opened 2019-06-29
NodeJS support

michelvocks:
Fixes #183 

Docs PR: https://github.com/gaia-pipeline/docs/pull/12

### DIFF
--- a/workers/pipeline/build_nodejs.go
+++ b/workers/pipeline/build_nodejs.go
@@ -12,6 +12,8 @@ import (
 	uuid "github.com/satori/go.uuid"
 )
 
+const nodeJSInternalCloneFolder  = "jsclone"
+
 // BuildPipelineNodeJS is the real implementation of BuildPipeline for NodeJS
 type BuildPipelineNodeJS struct {
 	Type gaia.PipelineType
@@ -23,7 +25,7 @@ func (b *BuildPipelineNodeJS) PrepareEnvironment(p *gaia.CreatePipeline) error {
 	uuid := uuid.Must(uuid.NewV4(), nil)
 
 	// Create local temp folder for clone
-	cloneFolder := filepath.Join(gaia.Cfg.HomePath, gaia.TmpFolder, gaia.TmpNodeJSFolder, srcFolder, uuid.String())
+	cloneFolder := filepath.Join(gaia.Cfg.HomePath, gaia.TmpFolder, gaia.TmpNodeJSFolder, srcFolder, uuid.String(), nodeJSInternalCloneFolder)
 	err := os.MkdirAll(cloneFolder, 0700)
 	if err != nil {
 		return err
@@ -56,7 +58,7 @@ func (b *BuildPipelineNodeJS) ExecuteBuild(p *gaia.CreatePipeline) error {
 	// Set command args for archive process
 	pipelineFileName := appendTypeToName(p.Pipeline.Name, p.Pipeline.Type)
 	args := []string{
-		"--exclude='./.git'",
+		"--exclude=.git",
 		"-czvf",
 		pipelineFileName,
 		"-C",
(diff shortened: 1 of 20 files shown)


workers/pipeline/update_pipeline_test.go
```diff
@@ -88,3 +88,40 @@ func TestUpdatePipelineRuby(t *testing.T) {
...
+	p1 := gaia.Pipeline{
+		Name:    "PipelinA",
+		Type:    gaia.PTypeNodeJS,
+		Created: time.Now(),
+	}
+
+	// Create fake test nodejs archive file.
+	src := filepath.Join(tmp, "PipelineA_nodejs")
+	p1.ExecPath = src
+	ioutil.WriteFile(src, []byte("testcontent"), 0666)
```
golangcibot: Error return value of `ioutil.WriteFile` is not checked (from `errcheck`)

workers/pipeline/update_pipeline.go
```diff
@@ -79,9 +85,45 @@ func updatePipeline(p *gaia.Pipeline) error {
 		// Install gem forcefully.
 		cmd := exec.Command(path, "install", "-f", pipelineCopyPath)
 		if out, err := cmd.CombinedOutput(); err != nil {
-			gaia.Cfg.Logger.Error("error", string(out[:]))
+			return fmt.Errorf("cannot install ruby gem: %s", string(out[:]))
+		}
+	case gaia.PTypeNodeJS:
```
Skarlso: At some point we'll need to refactor this into something sensible like a plugin type factory of some sorts, because this switch case is getting ridicules. :D

michelvocks: Yeah, I completely agree. 

I personally think for now it's "fine" since we have two or three switches which you need to modify when you add a new language so that's overall not much but for the future, a lot of refactoring will be required!

But that is a problem for the future Michel (😅)

Skarlso: :D :D :D Fine. :D

Skarlso:
Going to give this a spin now.

workers/pipeline/build_nodejs.go
```diff
@@ -0,0 +1,115 @@
...
+
+// BuildPipelineNodeJS is the real implementation of BuildPipeline for NodeJS
+type BuildPipelineNodeJS struct {
+	Type gaia.PipelineType
+}
+
+// PrepareEnvironment prepares the environment before we start the build process.
+func (b *BuildPipelineNodeJS) PrepareEnvironment(p *gaia.CreatePipeline) error {
+	// create uuid for destination folder
+	uuid := uuid.Must(uuid.NewV4(), nil)
```
Skarlso: Shouldn't we generalise to use `GenerateRandomUUIDV5`?

workers/pipeline/update_pipeline.go
```diff
@@ -79,9 +85,45 @@ func updatePipeline(p *gaia.Pipeline) error {
...
+		tmpFolder := filepath.Join(gaia.Cfg.HomePath, gaia.TmpFolder, gaia.TmpNodeJSFolder, p.Name)
+		os.RemoveAll(tmpFolder)
+
+		// Recreate the temp folder
+		if err := os.MkdirAll(tmpFolder, 0700); err != nil {
+			return err
+		}
+
+		// Unpack it
+		cmd := exec.Command(path, "-xzvf", p.ExecPath, "-C", tmpFolder)
```
Skarlso: I'm wondering if we really need the `-v` flag? I don't think we really need this to be verbose, don't we?

workers/pipeline/build_nodejs.go
michelvocks: Good idea. We can do that after merge and then for all builders?

Skarlso: Sure. Create a ticket. :D

workers/pipeline/update_pipeline.go
michelvocks: If the command fails, the output is displayed in the logs. This could be helpful.

PR #186, opened 2019-07-03
Upgrade of all frontend components and dependency libraries

michelvocks:


Skarlso:
Gotta love updates
```
Error: webpack.optimize.UglifyJsPlugin has been removed, please use config.optimization.minimize instead.
```
:D

michelvocks:
Alt Text

Skarlso:
Oh Gods :D

michelvocks:
@Skarlso This is now ready for review. I did a complete overhaul of the frontend development system. Instead of webpack dev environment, we are now using the vue-cli which is way better in my opinion.

michelvocks:
One thing to add. Instead of `npm run dev` you have to run now `npm run serve` to run the development server from the frontend. Will add this to the README file.

Skarlso:
@michelvocks Cool! Will do a test run of it. :)

Skarlso:
so two interesting notes... 

First:
If you now start the frontend first, it will bind to 8080 which will block the local gaia 8080. This is interesting because the previous one didn't do that.

Edit: Oh:
```
module.exports = {
  devServer: {
    proxy: {
      '^/api/v1': {
        target: 'http://localhost:8080',
        changeOrigin: false
      }
    }
  }
}
```

This will mess things up since Gaia backend default is 8080 too.

Second:

Why does it look so ugly now? :DDDD

Screenshot 2019-07-15 at 14 43 35

It got really black.

michelvocks:
> First:
If you now start the frontend first, it will bind to 8080 which will block the local gaia 8080. This is interesting because the previous one didn't do that.

Thanks. Forgot to add the default port to the dev server. Fixed that.

> Second:
Why does it look so ugly now? :DDDD

I've spent the most time to update the used table library since they released a new major version.
I decided to stick to the same style we already use for our gadgets like input, textarea and other gadgets. I personally like it more than the old one 😄

Skarlso:
@michelvocks It's a lot darker. I preferred the other one because it was easier on the eyes. :) But 🤷‍♂ :D

michelvocks:
Hmm. Maybe the text color is too dark? Let me try to make that brighter.

Skarlso:
Nono, the background.

Skarlso:
Oh, I see you already added the readme. :) Noice. I think this can be merged.

Issue #188, opened 2019-07-10
Pipeline Schedule button errors out


Skarlso:
After the latest update, the pipeline schedule button on Create Pipeline view is not working.

```
create.vue?41b9:665 Uncaught TypeError: Cannot read property 'length' of undefined
    at VueComponent.showPeriodicalPipelineScheduleModal (create.vue?41b9:665)
    at boundFn (vue.runtime.esm.js?a427:190)
    at invoker (vue.runtime.esm.js?a427:2000)
    at HTMLAnchorElement.fn._withTask.fn._withTask (vue.runtime.esm.js?a427:1798)
```

I'll fix this ASAP.

PR #189, opened 2019-07-10
Fixed schedule button

Skarlso:


Skarlso:
@michelvocks Hey man, I know you are away, but this needed an urgent fix. The schedule button wasn't working because the inner property could be undefined. I fixed it with a check. :)

Skarlso:
Closes #189.

Skarlso:
Closes #188

PR #191, opened 2019-07-12
Housekeeping!

Skarlso:
@michelvocks You are going to love this one... :D

### DIFF
--- a/handlers/hook.go
+++ b/handlers/hook.go
@@ -41,7 +41,7 @@ type Payload struct {
 
 func signBody(secret, body []byte) []byte {
 	computed := hmac.New(sha1.New, secret)
-	computed.Write(body)
+	_, _ = computed.Write(body)
 	return []byte(computed.Sum(nil))
 }
 
@@ -54,7 +54,7 @@ func verifySignature(secret []byte, signature string, body []byte) bool {
 	}
 
 	actual := make([]byte, 20)
-	hex.Decode(actual, []byte(signature[5:]))
+	_, _ = hex.Decode(actual, []byte(signature[5:]))
 	expected := signBody(secret, body)
 	return hmac.Equal(expected, actual)
 }
--- a/handlers/pipeline.go
+++ b/handlers/pipeline.go
@@ -1,10 +1,11 @@
 package handlers
 
 import (
-	"github.com/gaia-pipeline/gaia/helper/stringhelper"
 	"net/http"
 	"strconv"
 	"time"
+
+	"github.com/gaia-pipeline/gaia/helper/stringhelper"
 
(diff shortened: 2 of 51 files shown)


handlers/settings.go
```diff
@@ -27,13 +27,13 @@ func SettingsPollOn(c echo.Context) error {
 	gaia.Cfg.Poll = true
 	err = pipeline.StartPoller()
 	if err != nil {
-		c.String(http.StatusBadRequest, err.Error())
+		return c.String(http.StatusBadRequest, err.Error())
```
Skarlso: @michelvocks We had a couple of these. :DDDD we weren't returning errors man. :D

handlers/hook.go
```diff
@@ -41,7 +41,7 @@ type Payload struct {
 
 func signBody(secret, body []byte) []byte {
 	computed := hmac.New(sha1.New, secret)
-	computed.Write(body)
+	_, _ = computed.Write(body)
```
michelvocks: I'm wondering why we explicitly ignore errors here instead of doing what is recommended - checking the error and fail if there is one? 😅

handlers/pipeline.go
```diff
@@ -1,11 +1,12 @@
 package handlers
 
 import (
-	"github.com/gaia-pipeline/gaia/helper/stringhelper"
 	"net/http"
 	"strconv"
 	"time"
 
+	"github.com/gaia-pipeline/gaia/helper/stringhelper"
+
```
michelvocks: I don't know what is wrong with `go-import` (this always happens for me too) but this looks really ugly 😒

handlers/settings.go
michelvocks: 🙈 oh no. Good catch!

workers/agent/agent_test.go
```diff
@@ -18,6 +17,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/pkg/errors"
+
```
michelvocks: 😒

michelvocks (commented):
Nice! Thanks @Skarlso 
Wondering if we should try to check errors in general instead of explicitly ignoring them?

workers/agent/agent_test.go
Skarlso: hhehehehhe :D

Skarlso:
@michelvocks I have the handling in a different PR. This is part 1 basically. The handling and propagations of some of these errors when I was not ignoring them was actually causing some test failures.

And thus require a lot more thought and work to have them right. So I wanted to do this in one patch and then go over all the `_`-s and handle them if that's okay with you.

michelvocks:
Gotcha! Cool. Then LGTM 😄

PR #198, opened 2019-07-18
Refactored Gaia parameters

michelvocks:
This PR refactors Gaia's configuration parameters which had an inconsistent naming scheme and missed some important documentation.


docker/Dockerfile
```diff
@@ -307,7 +306,7 @@ RUN chmod +x ./gaia-linux-amd64 \
     && chmod 600 /root/.ssh
 
 # Set homepath as volume
-VOLUME [ "${GAIA_HOMEPATH}" ]
+VOLUME [ "${GAIA_HOME-PATH}" ]
```
Skarlso: `-` in environment properties is not a good thing as they usually can't be parsed nicely or cause problems on the command line in ZSH they even have to be explicitly escaped sometimes like `[` or `{`. I would just leave this out. It's fine like that.

michelvocks: We need some way to separate longer property options. My first idea was to use `_` but if you pass parameters to the binary it also looks odd: `-home_path`

Skarlso: Oh hang on that stuff is auto-read. ARGH. Okay, never mind then.
Also, yeah it should be `--home-path`

michelvocks: I found out that `flag` supports that out of the box by replacing `-` with `_` for environment variables. Perfect 😄

Skarlso: Oh very awesome! :D :D Perfect. :)

PR #200, opened 2019-07-25
Worker builds binary in case the binary doesn't match the worker arch

Skarlso:
@michelvocks Alright! This works, but, as you can see I had to export two of the none exported functions, NewBuild and GitPull. I can't use the Create one directly because that tries to store things in the db to which the worker doesn't have access to.

I'll try to refactor this, I just wanted to share with you the current progress.


workers/agent/agent.go
```diff
@@ -454,9 +455,76 @@ func (a *Agent) scheduleWork() {
 
 		// Let us try to start the plugin and receive all implemented jobs
 		if err = a.scheduler.SetPipelineJobs(pipeline); err != nil {
-			gaia.Cfg.Logger.Error("cannot get pipeline jobs", "error", err.Error(), "pipelinerun", pipelineRunPB)
-			reschedulePipeline()
-			return
+			if strings.Contains(err.Error(), "exec format error") {
```
michelvocks: It makes the code more readable if you do a negative check beforehand. e.g.:
```
if !strings.Contains(err.Error(), "exec format error") {
  gaia.Cfg.Logger.Error("cannot get pipeline jobs", "error", err.Error(), "pipelinerun", pipelineRunPB)
   reschedulePipeline()
   return
}
```

workers/agent/agent.go
```diff
@@ -454,9 +455,76 @@ func (a *Agent) scheduleWork() {
...
+				gaia.Cfg.Logger.Info("pipeline in a different format than worker; attempting to rebuild...")
+				err = nil
```
michelvocks: I think this is not necessary

workers/agent/agent.go
```diff
@@ -454,9 +455,76 @@ func (a *Agent) scheduleWork() {
...
+				gitRepo.SelectedBranch = repo.SelectedBranch
+				pCreate.Pipeline.Repo = &gitRepo
+				//gp.CreatePipeline(pCreate)
+				//if pCreate.StatusType == gaia.CreatePipelineFailed {
+				//	gaia.Cfg.Logger.Error("cannot create pipeline", "output", pCreate.Output, "pipelinerun", pipelineRunPB)
+				//	reschedulePipeline()
+				//	return
+				//}
+
+				pb := gp.NewBuildPipeline(pipeline.Type)
```
michelvocks: You can actually use `CreatePipeline` here. Worker also uses the storage service to store the pipeline run status updates.

workers/proto/worker.pb.go
```diff
@@ -3,13 +3,14 @@
 
 package protobuf
 
+import proto "github.com/golang/protobuf/proto"
```
michelvocks: This looks very different from my generation output. Which protoc version do you use? I use 3.7.0

workers/server/worker.go
```diff
@@ -102,6 +102,36 @@ func (w *WorkServer) GetWork(workInst *pb.WorkerInstance, serv pb.Worker_GetWork
 	return nil
 }
 
+// GetGitRepo checks out the code for a git repository.
+func (w *WorkServer) GetGitRepo(ctx context.Context, in *pb.PipelineID) (*pb.GitRepo, error) {
+	repo := &pb.GitRepo{}
+	// Check if worker is registered
```
michelvocks: Would love to see some blanket lines between code blocks to make it more readable 😄

Skarlso: Don't worry dude, I will make this pretty. :D

workers/agent/agent.go
Skarlso: I don't intend to actually keep this in this way. :D

workers/agent/agent.go
Skarlso: yep :)

workers/agent/agent.go
Skarlso: Interesting. I had an error when I tried that which said, nope. :O Huh. It might have been another problem. That would be cool, cause I think CreatePipeline does the gitCheckout and I wouldn't have to export these two functions.

workers/proto/worker.pb.go
Skarlso: ❯ protoc --version
libprotoc 3.7.1

And whatever the latest is from protoc-gen-go.

workers/agent/agent.go
Skarlso: Ah you can actually see me trying out CreatePipeline below which is commented out because it did not work. I'll look into it some more.

Skarlso:
@michelvocks So this works, however. On the previous iteration I wasn't saving the pipeline as a new create run. Now I am with create. On the previous iteration this lead to a nice thing where it didn't rebuild the pipeline again once the worker rebuild it for itself. Now it's always rebuilding and I'm not exactly sure why. Probably because I'm always creating a new `CreatePipeline` entry?

I also refactored a code a bit. ;)

go.sum
```diff
@@ -30,6 +30,8 @@ github.com/flynn/go-shlex v0.0.0-20150515145356-3f9db97f8568 h1:BHsljHzVlRcyQhjr
 github.com/flynn/go-shlex v0.0.0-20150515145356-3f9db97f8568/go.mod h1:xEzjJPgXI435gkrCt3MPfRiAkVrwSbHsst4LCFVfpJc=
 github.com/gaia-pipeline/flag v1.7.4-pre h1:/TAmHVYVQGE4Mw9xl0Qs0D5UruVDMF95thexyEFbTAY=
 github.com/gaia-pipeline/flag v1.7.4-pre/go.mod h1:rLpsWzqOEPa2K0Yl4aC34nmblLpIYjGqjP/srZbYvEk=
+github.com/gaia-pipeline/gosdk v0.0.0-20180909192508-cc9f89055777 h1:Gn3nmETr4IE44pIFLeTDNNBBRhNXXh53PqFV23CsVok=
+github.com/gaia-pipeline/gosdk v0.0.0-20180909192508-cc9f89055777/go.mod h1:e3TkvPdcdSdHZTgwiS89fs8lJrveYHsGLlz/Q0oXlN8=
```
Skarlso: uh?

go.mod
```diff
@@ -39,6 +40,7 @@ require (
 	github.com/stretchr/testify v1.3.0 // indirect
 	github.com/xanzy/ssh-agent v0.2.0 // indirect
 	golang.org/x/crypto v0.0.0-20190211182817-74369b46fc67
+	golang.org/x/net v0.0.0-20180826012351-8a410e7b638d
```
Skarlso: what the hel. the protobuf file is using /x/net? :O

workers/proto/worker.pb.go
```diff
@@ -21,7 +22,7 @@ var _ = math.Inf
 // is compatible with the proto package it is being compiled against.
 // A compilation error at this line likely means your copy of the
 // proto package needs to be updated.
-const _ = proto.ProtoPackageIsVersion3 // please upgrade the proto package
+const _ = proto.ProtoPackageIsVersion2 // please upgrade the proto package
```
Skarlso: This definitely looks older.

Skarlso: Fixed it. Turns out my plugin was an older plugin. That wasn't updated....

go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/emirpasic/gods v1.9.0 // indirect
 	github.com/flynn/go-shlex v0.0.0-20150515145356-3f9db97f8568 // indirect
 	github.com/gaia-pipeline/flag v1.7.4-pre
+	github.com/gaia-pipeline/gosdk v0.0.0-20180909192508-cc9f89055777
```
michelvocks: go mod automatically adds them when you test a pipeline build because the source code from the pipeline is in the tmp folder

Skarlso: interesting, I thought I made it ignore that folder. :D But apparently not. :)

michelvocks:
> @michelvocks So this works, however. On the previous iteration I wasn't saving the pipeline as a new create run. Now I am with create. On the previous iteration this lead to a nice thing where it didn't rebuild the pipeline again once the worker rebuild it for itself. Now it's always rebuilding and I'm not exactly sure why. Probably because I'm always creating a new `CreatePipeline` entry?
> 
> I also refactored a code a bit. ;)

I think the reason for the always rebuild is this line: https://github.com/gaia-pipeline/gaia/blob/master/workers/agent/agent.go#L403
You rebuild the binary locally and the next time when the pipeline is scheduled, the SHASum is different from the original pipeline binary.

We use the SHASum to detect if a pipeline has been changed which is kinda a problem here since we cannot detect if the pipeline has been changed. The only solution which came to my mind is to store the original SHASum before local rebuild on the worker side and on the next pipeline run (before rebuild) check if the original SHASum is still identical.

Skarlso:
@michelvocks Hang on. I would agree with this if it didn't work previously. But it isn't working while I call Create but was working when I skipped the storing of the new pipeline and was just simply rebuilding the binary. The SHA wouldn't have been the same in that case either, not?

michelvocks:
> @michelvocks Hang on. I would agree with this if it didn't work previously. But it isn't working while I call Create but was working when I skipped the storing of the new pipeline and was just simply rebuilding the binary. The SHA wouldn't have been the same in that case either, not?

I cannot remember how the code exactly looked when you did it this way 😄 I guess you forgot to update the SHASum? 😅

Skarlso:
Very good point. :D I did, yes. :D

Skarlso:
Okay, I'll save the original for a pipelinerun in the MemdB

Skarlso:
@michelvocks sorry, I just saw your edit question. :)

I have two rpi model 4s at home with 4Gig ram which I'm using as a Gaia worker. They both are linux. But my main laptop is a mac. So Gaia master is running on my mac and the worker is on ubuntu.

Skarlso:
you could also just use a VM actually. :D

Skarlso:
@michelvocks I'm not so sure about this. :D If I save the original SHA, and one comes in, and I only compare that, the rebuilt binary might be different not only because the worker rebuilt it, but also because of someone else slipping in a different binary which could be executed. I don't know man, sounds like a security risk.

What I could do, is save the original SHA, compare that, but also save the pipeline's new, rebuilt binary's SHA and also save that and compare. What do you think?

Skarlso:
Yeah, I definitely need two. One of the incoming one if that equals with the original, and that of the local one if that equals of the last one, because we want to make sure that the local binary has not been hijacked and altered.

Skarlso:
Okay, this actually works now, however two bugs, and todos:

* Check if there is actually a sha pair, I'm not checking for zero values right now
* The first run for some reason isn't working. It just returns with an empty build success but no job runs. Weird.
* This is gonna be a b*tch to unit test.

workers/agent/agent_test.go
```diff
@@ -363,6 +376,101 @@ func bufDialer(string, time.Duration) (net.Conn, error) {
...
+	// Run mScheduler
+	ag.scheduleWork()
+
+	// Validate output from mScheduler
+	if mStore.run == nil {
+		t.Fatal("run is nil but should exist")
+	}
+}
+
+func testRebuildWorkerBinary(t *testing.T) {
```
golangcibot: `testRebuildWorkerBinary` is unused (from `deadcode`)

Skarlso: I know. It creates a gaia.db for some reason. I haven't figured out yet why as everything should be on temp path.

Skarlso:
This f*cker runs for 44 seconds.

Skarlso:
@michelvocks Alright, this is as ready as it gets. I'll add some more coverage later on. Tell me if you like the SHAPair approach. :)

store/memdb/memdb.go
```diff
@@ -333,3 +343,63 @@ RunLoop:
...
+		return err
+	}
+
+	// Commit transaction
+	txn.Commit()
+	return nil
+}
+
+// GetSHAPair returns a sha pair for a pipeline id.
+func (m *MemDB) GetSHAPair(pipelineID string) (found bool, pair gaia.SHAPair, err error) {
```
michelvocks: Wouldn't it be easier to return a pointer and return a nil-pointer in case we couldn't find it?

workers/agent/agent.go
```diff
@@ -400,31 +407,42 @@ func (a *Agent) scheduleWork() {
 			reschedulePipeline()
 			return
 		}
+
+		// Let's check if we have a stored sha pair for this pipeline.
+		db, err := services.MemDBService(nil)
```
michelvocks: Comment above doesn't match what you actually do here

store/memdb/memdb.go
Skarlso: There is no point in sharing this struct, so, no. :) I don't see a "point" :D in returning a pointer. I would rather use the `bool, struct, error` pattern in this case. The less pointers the application is using the better the GC will be.

workers/agent/agent.go
```diff
@@ -454,12 +472,54 @@ func (a *Agent) scheduleWork() {
...
+			}
+
+			uID := strconv.Itoa(pipelineRun.PipelineID)
+			shaPair := gaia.SHAPair{
+				Original: pipelineSHA256SUM,
+				Worker:   workerSHA256Sum,
+				UniqueID: uID,
+			}
+
+			err = db.UpsertSHAPair(shaPair)
```
michelvocks: MemDB is only in-memory. What happens when a worker has been restarted?

store/memdb/memdb.go
michelvocks: When we load the data from the memdb we create a new struct anyway. The source routine which is calling this function is the only function which will own this struct. So this would actually increase the performance since you don't have to copy again the struct, right?

Skarlso: As long as I don't see benchmarks for that, that's an assumption. :)

If there is no point in sharing a data variable I won't share it. If I don't know what a copy does to a struct I will use a pointer, otherwise, value semantics should also be a first choice. Especially not up the call stack.

workers/agent/agent.go
Skarlso: Good question. It will disappear and would have to be added again. So if a worker stops it will re-compile everything after startup. Do you think that's reasonable thing to do? Or should we rather put things into the master's db?

michelvocks: > Or should we rather put things into the master's db?

No, that is not an option. But we could put it in the worker db, couldn't we? 😄

store/memdb/memdb.go
michelvocks: To be honest, I'm still confused about best practices around value/pointer usage. Some people say you should usually use values others say use whatever you want and then others say just be strict within your once code-base decision. 😞
If you look at the go standard libraries it is also mixed up quite heavy.

Skarlso: I think the standards are: 
- use value by default because it's safe and helps out the GC
- use pointers for sharing ( but only downwards avoid upwards sharing )
- use pointers if you definitely have a case where it provides a performance boost
- use pointers if you aren't sure if a value can be copied ( for example an unknown struct type or a reference / interface type ).

workers/agent/agent.go
Skarlso: @michelvocks Wait, uh... um... that's what I thought the memdb is? I might be confused here a bit. :D Would you mind clearing this up? Which one is the worker db? :D

michelvocks: We also use the boltdb (storage) on the worker side. MemDB is just a cache right before the storage to speedup certain things

Skarlso: Ah! Gotcha. Okay. I'll switch to boltdb. :)

Skarlso:
Buuu 

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x78 pc=0x10b10eb]

goroutine 246 [running]:
github.com/gaia-pipeline/gaia/handlers.(*mockStorageService).PipelineGetScheduled(0xc0002ec000, 0x32, 0xd65c103098ff4d62, 0x49b949fa3e89058a, 0x64e2d54e4c776eb3, 0x729f546e92990536, 0x542df71cb26c69b0)
	<autogenerated>:1 +0x6b
github.com/gaia-pipeline/gaia/workers/scheduler.(*Scheduler).schedule(0xc0000dca20)
	/go/src/github.com/gaia-pipeline/gaia/workers/scheduler/scheduler.go:234 +0xd1
github.com/gaia-pipeline/gaia/workers/scheduler.(*Scheduler).Init.func1(0xc0003ec870, 0xc0000dca20)
	/go/src/github.com/gaia-pipeline/gaia/workers/scheduler/scheduler.go:126 +0x4d
created by github.com/gaia-pipeline/gaia/workers/scheduler.(*Scheduler).Init
	/go/src/github.com/gaia-pipeline/gaia/workers/scheduler/scheduler.go:121 +0xdd
FAIL	github.com/gaia-pipeline/gaia/handlers	7.157s
Exited with code 1
```

:(

I'll try to take a look at this and fix it. It's bat that it can do that.

workers/agent/agent.go
Skarlso: Oh dang. I see now. This will make things a bit easier actually and maybe remove some added code. :)

store/store.go
```diff
@@ -120,54 +125,48 @@ func (s *BoltStore) Close() error {
...
+		bs:  s,
+		err: nil,
 	}
-	bucketName = workerBucket
-	err = s.db.Update(c)
-	if err != nil {
-		return err
+
+	// Make sure buckets exist
+	setP.update(userBucket)
```
Skarlso: @michelvocks Sorry, I had to refactor this in this batch. It was too repetitive. :)

michelvocks: 😄 good job!

Skarlso:
# Testing

Vagrant: `vagrant init ubuntu/bionic64`.
vim Vagrantfile -> uncomment public_network.
`vagrant up`
git clone gaia
checkout the pr
get a gaia config which creates a worker that points to the host machines IP address
run it -- it should connect
create a pipeline and run that pipeline
worker should rebuild the pipeline and execute it successfully
run it again
worker should just run what it already has without rebuilding
change the pipeline
worker should rebuild again since the original pipeline changed

store/sha_pair.go
```diff
@@ -0,0 +1,51 @@
...
+	})
+}
+
+// GetSHAPair returns a pair of shas for this pipeline run.
+func (s *BoltStore) GetSHAPair(pipelineID int) (ok bool, pair gaia.SHAPair, err error) {
+	return ok, pair, s.db.View(func(tx *bolt.Tx) error {
+		// Get bucket
+		b := tx.Bucket(shaPairBucket)
+
+		// Get pipeline
```
michelvocks: Wrong comment

store/sha_pair.go
```diff
@@ -0,0 +1,51 @@
...
+		v := b.Get(itob(pipelineID))
+
+		// Check if we found the pipeline
```
michelvocks: Wrong comment

store/sha_pair.go
```diff
@@ -0,0 +1,51 @@
...
+		if v == nil {
+			ok = false
+			return nil
+		}
+
+		// Unmarshal pipeline object
```
michelvocks: Wrong comment

michelvocks (approved):
LGTM ❤️

PR #201, opened 2019-08-01
Docker support for pipeline runs

michelvocks:


server/server.go
```diff
@@ -166,90 +174,92 @@ func Start() (err error) {
...
-		// Start worker gRPC server
-		workerServer := server.InitWorkerServer()
-		go workerServer.Start()
-
-		// Start listen
+		// Start API server
 		echoInstance.Logger.Fatal(echoInstance.Start(":" + gaia.Cfg.ListenPort))
 	case gaia.ModeWorker:
+		// Start API server
+		go echoInstance.Start(":" + gaia.Cfg.ListenPort)
```
golangcibot: Error return value of `echoInstance.Start` is not checked (from `errcheck`)

golangcibot: Error return value of `echoInstance.Start` is not checked (from `errcheck`)

workers/server/worker.go
```diff
@@ -78,14 +83,41 @@ func (w *WorkServer) GetWork(workInst *pb.WorkerInstance, serv pb.Worker_GetWork
...
+			}
+		case gaia.ModeWorker:
+			// Get information directly from the storage
+			p, err := store.PipelineGet(scheduled.PipelineID)
+			if err != nil {
+				gaia.Cfg.Logger.Error("failed to get pipeline via GetWork", "error", err.Error(), "pipeline", scheduled)
+				return err
 			}
+			if p == nil {
+				gaia.Cfg.Logger.Error("failed to find related pipeline via GetWork", "error", err.Error(), "pipeline", scheduled)
```
golangcibot: nilness: nil dereference in dynamic method call (from `govet`)

michelvocks:
@Skarlso This is ready for review. I still have to work on the documentation part which might be helpful for the review 😄

helper/pipelinehelper/pipelinehelper.go
```diff
@@ -12,4 +14,10 @@ const (
 // GetRealPipelineName removes the suffix from the pipeline name.
 func GetRealPipelineName(name string, pType gaia.PipelineType) string {
 	return strings.TrimSuffix(name, typeDelimiter+pType.String())
-}
\ No newline at end of file
+}
+
+// AppendTypeToName appends the type to the output binary name.
+// This allows later to define the pipeline type by the pipeline binary name.
+func AppendTypeToName(n string, pType gaia.PipelineType) string {
```
Skarlso: Can we try eliminating these helpers please or put them where they belong? The helper/stuff/util packages are a architectural clutch/anti-pattern. It's code of which no-one knows where to put or was put there for rudimentary decoupling. I'm sure we can find a place for these to live. Maybe their own package something that makes sense at the first sight? Like I look at helpers and I will have no idea what code to expect in there. Where as if I would look at package pipeline I'll expect things relating to pipelines.

Skarlso:
Great job! I'll review it. 😊

helper/pipelinehelper/pipelinehelper.go
michelvocks: I personally like having a `helper` package and I disagree that this is an architectural clutch/anti-pattern. Often, you find yourself in a situation where you have a few small and simple functions which you need to use in several different packages. If you put them in the `pipeline` package, you face quickly circular dependency issues which is a problem.
We could move all packages inside the helper package to the top-level but this makes the package structure way messier. 

> I look at helpers and I will have no idea what code to expect in there

The package structure is usually helper/<package-name> (e.g. helper/stringhelper). I personally think this is self-explaining. 

> Where as if I would look at package pipeline I'll expect things relating to pipelines.

So you are not confused when the package is called `pipelines` but you are when it is called `pipelinehelper`? 😄 

I also made some good experiences with this structure in other big projects (Vault: https://github.com/hashicorp/vault/tree/master/helper).

Skarlso: Anyways. At least it's covered with tests.

server/server.go
```diff
@@ -166,90 +172,94 @@ func Start() (err error) {
...
+	}
+
 	// Initialize scheduler
 	scheduler, err := services.SchedulerService()
 	if err != nil {
 		return
 	}
 
+	// Start worker gRPC server.
+	// We need this in both modes (server and worker) for docker worker to run.
```
Skarlso: What's the reason for needing it in both?

store/memdb/memdb.go
```diff
@@ -333,3 +351,138 @@ RunLoop:
...
+		return dockerWorkers, nil
+	}
+
+	// Iterate through all items and add them
+	for {
+		item := iter.Next()
+
+		if item == nil {
+			break
+		}
```
Skarlso: Why not put this check into the for? 

```go
for item := iter.Next(); item != nil {
```

helper/pipelinehelper/pipelinehelper.go
michelvocks: > The stringhelper is a good example that that thing could easily live in its own package. In fact, you already did put it into it's own package. :D

That is my point. It is already in its own package. I don't want to move that package to the top-level. Imagine you are new to the project and you need to find out where everything exists. If we have 20+ folders in the top-level folder, it will be a huge mess. The helper package provides a little benefit which makes the project structure a bit cleaner. 

> Yeah I know other big projects use helper, and I will strongly advocate in those big projects against it as well. :D

I know there are some Go-veterans on the internet which have the same opinion. I totally understand that point but I found out that it is simply not practicable. 😄

server/server.go
michelvocks: Good question. Imagine you want to run pipelines on a remote Gaia worker. In this case the following will happen:

1. Worker retrieves pipeline run from Gaia primary instance.
2. Worker starts another worker in a docker container.
3. Worker in the docker container connects to the worker.
4. Worker in docker container executes pipeline.
5. Worker in docker container sends back run logs & status.
6. Worker sends back logs & status to primary instance.

store/memdb/memdb.go
michelvocks: Good catch! I will change that! 😄

server/server.go
Skarlso: Ahh I see. Makes sense.

Skarlso:
Sorry, only picking through this slowly. But I'll get there. :)

workers/scheduler/scheduler.go
```diff
@@ -275,18 +283,59 @@ func (s *Scheduler) schedule() {
...
 
-		// Update entry in store
-		err = s.storeService.PipelinePutRun(scheduled[id])
-		if err != nil {
-			gaia.Cfg.Logger.Debug("could not put pipeline run into store", "error", err.Error())
+			// Cache docker worker
+			if err := s.memDBService.InsertDockerWorker(worker); err != nil {
+				gaia.Cfg.Logger.Error("failed to cache docker worker in memdb", "error", err)
+				if err := worker.KillDockerWorker(); err != nil {
+					gaia.Cfg.Logger.Error("failed to kill docker worker", "error", err)
```
Skarlso: I feel like simply logging this fact is weird. At the least we should wrap it and pass it up, don't you think? Or provide some context, I don't know. If we just log it, I feel like we simply don't care and might as well not log it. Especially since Gaia is very verbose now and this might be just missed.

michelvocks: > At the least we should wrap it and pass it up, don't you think?

Can you provide an example? If you mean with "pass it up" returning the error then this is not possible. We are in the schedule function which is periodically triggered and doesn't allow a return parameter.

The reason why I added this output is only for debugging purposes. What would you expect to happen here?

michelvocks:
Related Docs PR: https://github.com/gaia-pipeline/docs/pull/14

workers/scheduler/scheduler.go
Skarlso: Yeah the scheduler never returns an error. I guess it's okay, it's already in a failing statement.

workers/docker/docker.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	// Create container
+	resp, err := createContainer()
+	if err != nil {
+		gaia.Cfg.Logger.Error("failed to create worker container", "error", err)
+		gaia.Cfg.Logger.Info("try to pull docker image and then try it again...")
+
+		// Pull worker image
+		_, err = cli.ImagePull(ctx, workerImage, types.ImagePullOptions{})
```
Skarlso: It's worth mentioning that right now this only supports public docker registry and no private registries. I guess we can add that functionality later on?

michelvocks: Good catch! Yes, exactly. I wouldn't prioritize this yet since we can easily add this at a later point in time.

helper/pipelinehelper/pipelinehelper.go
Skarlso: Hm. I guess I can see that in this case it's okay, since it's in its own package that makes sense. I gues the problem is when the helper is just a jumble of stuff extracted for the sake of extraction. What I don't want to see is code that has no purpose or exists in a utils package we don't know what it does or how it's used and the extraction only happened because we wanted to avoid to copy a bit of code, or to avoid loops in imports. :)

What I'm saying is that I can live with this. :DDDDD

workers/docker/docker.go
```diff
@@ -0,0 +1,132 @@
...
+
+	if resp.State.Running {
+		return true
+	}
+	return false
+}
+
+// KillDockerWorker disconnects the existing docker worker and
+// kills the docker container.
+func (w *Worker) KillDockerWorker() error {
```
Skarlso: Why not start the container with `--rm`?  I'm sure the api can do that?

Skarlso:
@michelvocks So I wanted to test this now, and I found two bugs... 

First, when I created the pipeline and enabled the docker thing on the pipeline creation ui, it wasn't enabled for the pipeline. The indicator on the pipeline view said disabled and in fact it did not run in a container.

Screenshot 2019-10-11 at 9 27 22

Once I enabled this though... It failed to run unfortunately by containously spewing this:

```
2019-10-11T09:29:30.006+0200 [INFO ] Gaia: try to pull docker image and then try it again...
2019-10-11T09:29:31.864+0200 [ERROR] Gaia: failed to create worker container after pull: error="Error response from daemon: No such image: gaiapipeline/gaia:latest"
2019-10-11T09:29:31.864+0200 [ERROR] Gaia: failed to setup docker worker for pipeline run: error="Error response from daemon: No such image: gaiapipeline/gaia:latest"
2019-10-11T09:29:33.007+0200 [ERROR] Gaia: failed to create worker container: error="Error response from daemon: No such image: gaiapipeline/gaia:latest"
2019-10-11T09:29:33.007+0200 [INFO ] Gaia: try to pull docker image and then try it again...
2019-10-11T09:29:35.262+0200 [ERROR] Gaia: failed to create worker container after pull: error="Error response from daemon: No such image: gaiapipeline/gaia:latest"
2019-10-11T09:29:35.262+0200 [ERROR] Gaia: failed to setup docker worker for pipeline run: error="Error response from daemon: No such image: gaiapipeline/gaia:latest"
```

Also, this made me realise that there is no timeout on this thing. :D This can keep failing and retrying indefinitely.

workers/docker/docker.go
michelvocks: As far as I know `--rm` does not work with detached containers (e.g. `-d`). Additionally, I have to stop the container anyway so that the container would be removed by the `--rm`.

michelvocks:
> First, when I created the pipeline and enabled the docker thing on the pipeline creation ui, it wasn't enabled for the pipeline. The indicator on the pipeline view said disabled and in fact it did not run in a container.

Let me try to reproduce this.

> Once I enabled this though... It failed to run unfortunately by containously spewing this:
> Not sure what's wrong because gaiapipeline/gaia:latest is an existing image. But it failed to pull it. I think it didn't block and wait for pulling the image?

I _should_ wait for the pull to finish but I will test that again. However, please keep in mind that we didn't release a new version (and also not a new docker image). That means the current Gaia version in the `latest` image has not the worker stuff included and will therefore not work. I manually build a image from master-branch and used that for testing.  

> Also, this made me realise that there is no timeout on this thing. :D This can keep failing and retrying indefinitely.

Right. I thought about that too and I personally think that we maybe don't want a time out? e.g. try it as long as possible strategy.

Skarlso:
The problem with `as long as possible` is that this continued on until I killed Gaia. Which means the pipeline is blocked until a Gaia restart occurs. I don't think that's a good idea.

True, the image wouldn't have the new worker stuff, but that's not the error. :) The error is that the image is not found. Which is weird because the image IS there. Did you try pushing your image up somewhere and deleting it locally and trying to run it like that?

store/memdb/memdb.go
michelvocks: That actually doesn't work. `item := iter.Next()` is only executed once before loop iteration.

michelvocks:
@Skarlso 

> First, when I created the pipeline and enabled the docker thing on the pipeline creation ui, it wasn't enabled for the pipeline. The indicator on the pipeline view said disabled and in fact it did not run in a container.

I tried to reproduce this but it somehow works for me. Can you just give it another try? Maybe you clicked in the create pipeline UI on the cancel button? 😄 

> True, the image wouldn't have the new worker stuff, but that's not the error. :) The error is that the image is not found. Which is weird because the image IS there. Did you try pushing your image up somewhere and deleting it locally and trying to run it like that?

There was a bug in the image pull code. Should now work 👍 

> The problem with as long as possible is that this continued on until I killed Gaia. Which means the pipeline is blocked until a Gaia restart occurs. I don't think that's a good idea.

I'm wondering if this is actually a problem in general and should be addressed in a different PR. I mean this can also happen for normal pipeline runs, right? I think we should add a "context" to a pipeline run with a timeout which makes sure after a certain time a pipeline is will be canceled?

Skarlso:
yo. :) I'll re-try as soon as possible! :)

Skarlso:
> I'm wondering if this is actually a problem in general and should be addressed in a different PR. I mean this can also happen for normal pipeline runs, right? I think we should add a "context" to a pipeline run with a timeout which makes sure after a certain time a pipeline will be canceled?

Agreed!

workers/server/server_test.go
```diff
@@ -0,0 +1,43 @@
...
+	gaia.Cfg.Logger = hclog.New(&hclog.LoggerOptions{
+		Level: hclog.Trace,
+		Name:  "Gaia",
+	})
+
+	// Init worker server
+	server := InitWorkerServer()
+
+	// Start server
+	go func() {
```
golangcibot: SA2002: the goroutine calls T.Fatal, which must be called in the same goroutine as the test (from `staticcheck`)

Skarlso:
going to try this now on a fresh laptop. :)

store/memdb/memdb.go
Skarlso: true I means to say `item := node; item != nil item = iter.Next`.

Skarlso:
eh, couldn't build the image...
```
E: Version '8u181-b13-2~deb9u1' for 'openjdk-8-jdk' was not found
E: Version '20170531+nmu1' for 'ca-certificates-java' was not found
```

Skarlso:
These new versions worked:

```
ENV JAVA_VERSION 8u232
ENV JAVA_DEBIAN_VERSION 8u232-b09-1~deb9u1

# see https://bugs.debian.org/775775
# and https://github.com/docker-library/java/issues/19#issuecomment-70546872
ENV CA_CERTIFICATES_JAVA_VERSION 20170929~deb9u3
```

How do you build your image?

Skarlso:
Okay, obviously you can't just build this thing. :D

```
COPY failed: stat /var/lib/docker/tmp/docker-builder354759785/docker/settings-docker.xml: no such file or directory
```

Skarlso:
Oh buuu 

```
Step 46/51 : COPY gaia-linux-amd64 /app
COPY failed: stat /var/lib/docker/tmp/docker-builder648944948/gaia-linux-amd64: no such file or directory
```

michelvocks:
It feels like every 2-3 month the newest Java version is already deprecated again and it is not possible to build the docker image again. I think we just have to regularly bump the version.

Usually, the docker image build process is started by goreleaser that is why it is expected to be in the root folder.

Skarlso:
@michelvocks I had a play with this last night and I feel like it's doing what it's supposed to do. :D The building is clunky, true. :/ I have yet to find a good solution for that. We can create a ticket for that maybe and track it?

So, LGTM! :)

Skarlso:
@michelvocks How about we ditch the pinning to a specific Java 8 open jdk version? I don't see any problems with trying to use whatever latest stable is in there, do you?

So this, works perfectly fine:

```
	apt-get install -y --no-install-recommends \
		openjdk-8-jdk \
		ca-certificates-java \
```

michelvocks:
@Skarlso Sounds good. I will do that in a separate PR.

Issue #204, opened 2019-08-19
There is no len check on active pipelines remove / update methods


Skarlso:
```go
// Update updates a pipeline at the given index with the given pipeline.
func (ap *ActivePipelines) Update(index int, p gaia.Pipeline) {
	ap.Lock()
	defer ap.Unlock()

	ap.Pipelines[index] = p
}

// Remove removes a pipeline at the given index from ActivePipelines.
func (ap *ActivePipelines) Remove(index int) {
	ap.Lock()
	defer ap.Unlock()

	ap.Pipelines = append(ap.Pipelines[:index], ap.Pipelines[index+1:]...)
}
```

These can and do cause index out of bounds errors.

PR #206, opened 2019-08-30
Created an error value, in case the for loop fails with index not found.

Skarlso:


workers/pipeline/pipeline.go
```diff
@@ -120,19 +120,28 @@ func (ap *ActivePipelines) Append(p gaia.Pipeline) {
...
 }
 
 // Remove removes a pipeline at the given index from ActivePipelines.
-func (ap *ActivePipelines) Remove(index int) {
+func (ap *ActivePipelines) Remove(index int) error {
 	ap.Lock()
 	defer ap.Unlock()
 
+	l := len(ap.Pipelines)
+	if index >= l || index+1 > l {
```
michelvocks:
```suggestion
	if index >= l || index < 0 {
```

workers/pipeline/pipeline.go
```diff
@@ -220,7 +231,10 @@ func (ap *ActivePipelines) RemoveDeletedPipelines(existingPipelineNames []string
 		}
 	}
 	for _, idx := range deletedPipelineIndices {
-		ap.Remove(idx)
+		if err := ap.Remove(idx); err != nil {
+			gaia.Cfg.Logger.Error("failed to remove pipeline with index: ", idx, "error", err.Error())
```
michelvocks:
```suggestion
			gaia.Cfg.Logger.Error("failed to remove pipeline with index", "index", idx, "error", err.Error())
```

michelvocks:
@Skarlso overall LGTM. Only these two issues

workers/pipeline/pipeline.go
Skarlso: Hey. :) 

So, yeah, I left out 0 that's a good call. However, I'll tell you why `index+1 > l` needs to be there. :D It's a mind bender so... 

This is the line we are protecting:
`ap.Pipelines = append(ap.Pipelines[:index], ap.Pipelines[index+1:]...)`
if l == 0:
The first part of append will fail.
if l == 1 and index = 1:
The first part will work because slice postfix is not inclusive.
But the second part, the `index+1` will fail because that one is inclusive. And 2 is not a valid index to look from.

Skarlso:
Closes #204.

Issue #205, opened 2019-08-25
It's high time. Do a trace and pprof run on Gaia


Skarlso:

Skarlso:
Actually, I know there are more immediate issues out there which need attention, but man I ran into a race condition with the tests again today. So I'm a bit afraid that we should optimise the routines a bit.

michelvocks:
Yeah, sounds like a good idea. Especially the scheduler is critical and could potentially have some leaking routines.

Skarlso:
Aye. I agree. I'll get into it, once the worker rebuild is finished. I'm really eager to trace things. :)

Skarlso:
Alright, found my first data race already. Going to work to fix this one first as it's the easiest to find and reproduce.

```
==================
WARNING: DATA RACE
Read at 0x00c0003a8768 by goroutine 97:
  bufio.(*Writer).Flush()
      /usr/local/Cellar/go/1.13/libexec/src/bufio/bufio.go:590 +0x79
  github.com/gaia-pipeline/gaia/plugin.(*GoPlugin).FlushLogs()
      /Users/hannibal/goprojects/gaia/gaia/plugin/plugin.go:322 +0x50
  github.com/gaia-pipeline/gaia/workers/scheduler.(*Scheduler).executeScheduler.func1()
      /Users/hannibal/goprojects/gaia/gaia/workers/scheduler/scheduler.go:556 +0x178

Previous write at 0x00c0003a8768 by goroutine 26:
  bufio.(*Writer).Write()
      /usr/local/Cellar/go/1.13/libexec/src/bufio/bufio.go:638 +0x415
  github.com/hashicorp/go-plugin.(*Client).logStderr()
      /Users/hannibal/golang/pkg/mod/github.com/hashicorp/go-plugin@v0.0.0-20180331002553-e8d22c780116/client.go:759 +0x321

Goroutine 97 (running) created at:
  github.com/gaia-pipeline/gaia/workers/scheduler.(*Scheduler).executeScheduler()
      /Users/hannibal/goprojects/gaia/gaia/workers/scheduler/scheduler.go:551 +0x393
  github.com/gaia-pipeline/gaia/workers/scheduler.(*Scheduler).executeScheduledJobs()
      /Users/hannibal/goprojects/gaia/gaia/workers/scheduler/scheduler.go:504 +0xd7
  github.com/gaia-pipeline/gaia/workers/scheduler.(*Scheduler).prepareAndExec()
[...]

PR #213, opened 2019-09-10
Created a concurrent safe Gaia logger for plugins.

Skarlso:
Deals with the first data race encountered during the investigating in #205.


Skarlso:
Also tested with workers.

Screenshot 2019-09-10 at 16 53 48
Screenshot 2019-09-10 at 16 53 44

Looks good! :)

PR #216, opened 2019-10-05
Fix issue #215, Allow spaces in Python pipeline name

### LINKED ISSUE #215: Bug: Cannot create Python pipelines with spaces in the pipeline name.
Pipeline creation fails each time when creating a Python pipeline with spaces in the pipeline name.
This is not the case when creating a Golang pipeline with spaces in the pipeline name, the Golang pipeline is created successfully.

## Steps to reproduce
1. Click on create new pipeline
2. Use the demo for git repo link `https://github.com/gaia-pipeline/python-example`
3. For the name of the pipeline give `Test Demo Pipeline`
4. Click on create pipeline 

## output 

```
cannot update pipeline: cannot install python plugin: DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support
ERROR: Invalid requirement: '/Users/sk/github/gaia/tmp/tmp/python/test'
Hint: It looks like a path. File '/Users/sk/github/gaia/tmp/tmp/python/test' does not exist.
```

## Expected output
Pipeline creation should be successful

If i change the Pipeline name to `TestDemoPipeline` (remove all the spaces), the pipeline is  created successfully.

Shivakishore14:
Fixes #215 
Wrap the argument for pip install with `'` so that `<spaces>` are escaped by shell.

Refer the issue for more info.

### DIFF
--- a/workers/pipeline/update_pipeline.go
+++ b/workers/pipeline/update_pipeline.go
@@ -3,10 +3,11 @@ package pipeline
 import (
 	"errors"
 	"fmt"
-	"github.com/gaia-pipeline/gaia/helper/filehelper"
 	"os"
 	"os/exec"
 	"path/filepath"
+
+	"github.com/gaia-pipeline/gaia/helper/filehelper"
 
 	"github.com/gaia-pipeline/gaia"
 )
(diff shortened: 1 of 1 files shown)


Skarlso:
Hello! Thank you for your contrib. Will take a look at this shortly.

PR #220, opened 2019-10-07
.sort causes a cascading update on the vue component

### LINKED ISSUE #212: Worker deregister button is broken
Currently on master branch, the settings -> worker -> deregister worker button does not work.

Skarlso:
@michelvocks I can't believe I finally found it. The sort I added on the tags? Apparently vue doesn't like the component being updated mid display. :/ So it does a re-eval on it, which causes another sort, which causes another re-eval... etc.

I have to have a different way of sorting the tags. I tried compute: but there was a different error. I'll make something up, but for now, I'm just removing the sort so everything works again.

BTW This is a new vue addition, the old one didn't do this. :D

Closes #212

### DIFF
--- a/frontend/src/views/settings/worker/manage-worker.vue
+++ b/frontend/src/views/settings/worker/manage-worker.vue
@@ -64,7 +64,7 @@
               </span>
             </span>
             <span v-if="props.column.field === 'tags'">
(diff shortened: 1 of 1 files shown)


michelvocks:
Nice! Good catch!

PR #221, opened 2019-10-07
[WIP] Fix: Python pipeline fails with non ascii chars in Pipeline name

### LINKED ISSUE #219: Python pipeline creation fails when pipeline name has non-ascii character
Steps to reproduce

1. Create new pipeline
2. For Repo, use the example repo `https://github.com/gaia-pipeline/python-example` 
3. For Pipeline name give `test文`

**Output**:
Pipeline creation fails

**Expected Output**:
Pipeline should be created successfully

The source of the issue can be traced to creation of the virtual environment. This is a known issue with virtualenv
https://github.com/pypa/virtualenv/issues/457

The pipeline also fails with the name that contains `$` followed by one or more characters
example :  `test$python`.

Shivakishore14:
Closes #219 
Use base64 encoded pipeline name for virtual env creation as it fails with non ascii chars.

Refer to issue for more information.

### DIFF
--- a/workers/pipeline/update_pipeline.go
+++ b/workers/pipeline/update_pipeline.go
@@ -1,6 +1,7 @@
 package pipeline
 
 import (
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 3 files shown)


Skarlso:
Hi @Shivakishore14! Looks good to me. Could you please do some manual testing and show the output? :) Cheers!

Shivakishore14:
@Skarlso There is an alternative solution suggested by @michelvocks , and that is a cleaner approach where users cannot enter these special characters which cause this issue.
https://github.com/gaia-pipeline/gaia/issues/219#issuecomment-539381657

Let me know if we can take that approach :)

Skarlso:
Yes, please take that approach and thank you for your patience!

Skarlso:
I will close this which has been fixed by restricting the pipeline name.

PR #222, opened 2019-10-07
A little refactor to remove the need for nil parameters.

Skarlso:


go.mod
```diff
@@ -48,3 +48,5 @@ require (
 	gopkg.in/warnings.v0 v0.1.2 // indirect
 	gopkg.in/yaml.v2 v2.2.2
 )
+
+go 1.13
```
michelvocks: Is that necessary? I still use 1.12.X since we had a lot of issues with 1.13 and go modules 😄

services/service_provider.go
```diff
@@ -116,10 +121,19 @@ func VaultService(vaultStore security.VaultStorer) (security.VaultAPI, error) {
 
 // MockVaultService provides a way to create and set a mock
 // for the internal vault service manager.
-func MockVaultService(service security.VaultAPI) {
+func MockVaultService(service security.GaiaVault) {
 	vaultService = service
 }
 
+// DefaultMemDBService provides a default memDBService with an underlying storere.
```
michelvocks:
```suggestion
// DefaultMemDBService provides a default memDBService with an underlying storer.
```

michelvocks (approved):
Overall LGTM! 😄

go.mod
Skarlso: No, I'm going to remove it. It's annoying that go 1.13 does this. :/

Skarlso: Also, yes. I'm probably going to revert to 1.12.X.

Skarlso:
\o/  
```
$ go version
go version go1.12.10 darwin/amd64
```

PR #229, opened 2019-12-04
Obfuscate execpath from pipeline objects returned to clients

### LINKED ISSUE #226: Stop sending pipeline local path to clients
Currently, we send all pipeline information to the clients (e.g. UI users). That also includes the local binary path of the pipeline from the Gaia server. This is confidental information and shouldn't be transferred to the clients.

michelvocks:
Fixes #226


Skarlso:
Hum, are none of the workers using that for anything?

michelvocks:
Workers don't use the usual REST API. They use the gRPC interface.

PR #237, opened 2020-02-23
Refactored scheduler out of the service singleton provider

Skarlso:


Skarlso:
@michelvocks Bask in this beauty! :)) Arguably, the naming could be better, also, I'm thinking of creating some more folders in order to separate the provider and the services better.

However, this works ( I can now see that some of the tests failed, although locally everything ran... grrr. ). It makes the tests a heck of a lot more stable because you don't have to worry about the fact that something overwrites your mock.

Eventually, if you like this way, I would like to refactor the rest of the singletons as well into services and providers.

services/service_provider.go
```diff
@@ -3,20 +3,20 @@ package services
...
-	"github.com/gaia-pipeline/gaia/workers/scheduler"
 )
 
 // storeService is an instance of store.
 // Use this to talk to the store.
 var storeService store.GaiaStore
 
 // schedulerService is an instance of scheduler.
-var schedulerService scheduler.GaiaScheduler
+var schedulerService gaiascheduler.Scheduler
```
golangcibot: `schedulerService` is unused (from `varcheck`)

Skarlso:
After master merge this isn't scheduling properly... hm 🤔

Skarlso:
This is now working, however I need to do some exhaustive testing around triggering, agent, docker runners and remote workers as well.

Skarlso:
Tested remote worker: ✅

PR #238, opened 2020-03-07
Output for jobs

Skarlso:
@michelvocks Hi! So this is ALMOST working. :D The output is saved, but I'm failing to figure out why the Job which is supposed to come next isn't seeing the depending on job's output.

I tried to update the Argument value in Execute and executeJob or something like that, but the depending job has actually an empty output there. Even though it's saved correctly now in PipelineRun. I'm stuck. :)


Skarlso:
@michelvocks Okay, so I found that the job isn't up to date with its dependencies that's why I can't find the output. But if I pass in the run to execute I can see the updated job outputs. Now, I'm not sure if this is right, please review and say if this isn't how its supposed to work. :)

Skarlso:
Hi @prologic.

So, the testing plan would be to pull this PR and build Gaia. With the respective binaries, you would want to test the output with something like this: https://github.com/Skarlso/go-example/tree/output_for_jobs

The important bit to note here is the fact that outputs only work with depending jobs. Also, if multiple outputs are provided by multiple depending jobs with the same key, only the first found will be returned. The rest is ignored. So keys need to be unique to work but I think that's fine.

You would want to create a pipeline and just run it. Test it with docker and test it with remote workers as well please, I will do the same. :) 

And thank you! Do you need more information? Usually how I run Gaia when I'm testing is just clone, and run `make`. Then go to the frontend and run `npm run serve`. If you want to test the worker on a remote machine you would run something like this: 

https://docs.gaia-pipeline.io/operations/worker/installation/

Please don't hesitate to ask if you got stuck or don't know something. I'll gladly help.

Cheers!

prologic:
Sorry haven't had a chance to test this out yet been rather busy with my real day job :) I will try to find some time to look at this soon!

Skarlso:
Thanks, no problem. I know that these are difficult times right now. :)

Skarlso:
I did some more testing on this and appears to be working with workers and remote things as well. Now just needs a proper review if the code change is acceptable by you @michelvocks :). 

@prologic Did you manage to do anything with it by any chance? It's okay if not, I'm just asking. :)

Cheers.

prologic:
> @prologic Did you manage to do anything with it by any chance? It's okay if not, I'm just asking. :)

Sadly no; Been so busy lately :/ Sorry!

Skarlso:
No problem my man. I'll do some testing later on. :)

PR #244, opened 2020-05-01
[WIP] Look at moving permissions onto the echo handlers (as middleware)

speza:
Now I have come back to this after a while out, I feel like having to define the exact API route of each permission is fragile and not very flexible.

With this change I have played around with moving the permissions (RBAC) onto the echo hanlders instead. While it pollutes the echo handlers a little bit more, its much more flexible since we don't have to manually write out the API endpoint into each permission role.


handlers/auth_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"crypto/rand"
 	"crypto/rsa"
 	"fmt"
+	"github.com/gaia-pipeline/gaia/helper/rolehelper"
```
Skarlso, line 7: Please keep the following import path ordering:
- stdlib
- third party libraries
- gaia outside imports
- gaia internal imports

All divided by blank lines in between the sections. :)

handlers/handler.go
```diff
@@ -1,6 +1,7 @@
 package handlers
 
 import (
+	"github.com/gaia-pipeline/gaia/helper/rolehelper"
```
Skarlso, line 4: Here too.

speza:
@Skarlso this is actually a great idea. I take it you're thinking similar to an AWS IAM role/policy?
Happy to look into a solution like that.

So, thinking about this... an action could be what we map in the code. This is where we could apply an 'action' check to the echo handlers to make sure a user has the required permission. This would obviously have to be something defined by us.

Its then a user defined policy/role that can include/exclude this actions?

Skarlso:
@speza Sounds about right. Keep in mind that Gaia is a long running process, so instead of making a lookup on every request, use a role cache on the name or something like that which we update every 5-10-15 minutes or so with a polling mechanism using the time.Ticker. :)

Yes the actions would have to be defined by us. Correct.

speza:
Closing. Working on a better solution :)

PR #246, opened 2020-05-08
[WIP] Resource yaml based rbac

speza:
Massive WIP.

Description TBD.

**Todo:**
- Refinement 
- Proper testing
- Cache eviction
- UI (another branch/PR maybe?)


handlers/auth.go
```diff
@@ -89,6 +101,17 @@ type AuthConfig struct {
 	RoleCategories []*gaia.UserRoleCategory
 }
 
+func getPoliciesFromClaims(policyClaims interface{}) []string {
+	policies := []string{}
+	if policyClaims == nil {
```
Skarlso, line 106: If this could potentially be nil inside the interface, this check will not be enough.

You'll have to also check the value in the interface to be nil or not using reflection.

```
if policyClaims == nil || reflect.ValueOf(policyClaims).IsNil() {
```

handlers/auth.go
```diff
@@ -89,6 +101,17 @@ type AuthConfig struct {
...
+		return policies
+	}
+	for _, p := range policyClaims.([]interface{}) {
```
Skarlso, line 109: Do a type assert here on `[]interface{}` to make sure it doesn't throw a type exception.

handlers/auth.go
```diff
@@ -89,6 +101,17 @@ type AuthConfig struct {
...
+		policies = append(policies, p.(string))
```
Skarlso, line 110: same here

handlers/auth.go
```diff
@@ -89,6 +101,17 @@ type AuthConfig struct {
 	RoleCategories []*gaia.UserRoleCategory
 }
 
+func getPoliciesFromClaims(policyClaims interface{}) []string {
+	policies := []string{}
```
Skarlso, line 105: In this case, I'd prefer a named return value for clarity. But I'll leave it up to you. :)

handlers/handler.go
```diff
@@ -2,6 +2,14 @@ package handlers
 
 import (
 	"net/http"
+	"time"
```
Skarlso, line 5: Please sort the includes according to ( in all other files too ):

1. stdlib
empty line
2. outside libraries
empty line
3. external gaia libraries
empty line
4. internal gaia libraries.

helper/cachehelper/cachehelper.go
```diff
@@ -0,0 +1,75 @@
+package cachehelper
+
+import (
+	"sync"
+	"time"
+)
+
+// Cache represents the interface for a simple cache.
+type Cache interface {
```
Skarlso, line 9: Since we know that these will be string values, I think this cache doesn't need to be interface valued, right? Unless I'm missing something and there could be other types values?

helper/cachehelper/cachehelper.go
```diff
@@ -0,0 +1,75 @@
...
+	Get(key string) (interface{}, bool)
+	Put(key string, value interface{}) interface{}
+}
+
+type cacheItem struct {
+	Value      interface{}
```
Skarlso, line 15: Why are these exported?

security/rbac/enforcer.go
```diff
@@ -0,0 +1,98 @@
...
+	return false
+}
+
+func (s *policyEnforcer) ResolvePolicies(policyNames []string) namespaceActionMap {
+	na := make(namespaceActionMap)
+
+	// iterate through the user policy names provided
+	for _, policy := range policyNames {
+		// get the policy from the service/cache
+		policyResource, _ := s.svc.GetPolicy(policy)
```
Skarlso, line 52: Do check for errors here though... If the policy doesn't exists, this should throw some kind of warning.

security/rbac/enforcer.go
```diff
@@ -0,0 +1,98 @@
...
+		// iterate through the policy statement contained in the retrieved policy
+		for _, stmt := range policyResource.Statement {
+			// iterate through the actions in the statement
+			for _, stmtAction := range stmt.Action {
+				// parse the namespace and action from the statement value
```
Skarlso, line 57: We should refactor this O(n^3)... :))) This should be very quick considering this will happen for every request.

handlers/handler.go
speza: Just need to sort my fmt out, np :)

helper/cachehelper/cachehelper.go
speza: I thought the cache could be generic enough that we could reuse if required?

security/rbac/enforcer.go
speza: Yeah agreed! This was the first basic impl. Now I have some passing tests I will refactor it for sure,

speza:
> First glance comments. I'll have to do another rounds for a deeper inspection. These are surface remarks for now. :) Generally, I saw what I expect I would see mostly, things I think are missing yet right? Like calling the cache eviction and the likes. Unless my glance missed that :)

Yep, correct mate. This was a rough PR that is definitely not ready. I just wanted some thoughts before I carried on. Its suprsingly a lot of code to get this system in! Especially to then refine it. I'd usually do this sort of work in many multiple PRs!

I left a short list of the required work still:
- Code refinement
- Testing
- Cache eviction and update when a resource has been modified  
- UI (another branch/PR maybe?)
- Think about how we handle default policies for existing users...

helper/cachehelper/cachehelper.go
Skarlso: we'll alter it when we get there I think. :) We might never reuse it. :)

If we would, it's very easy to change it to interface.

speza: Sure thing :+1:

handlers/handler.go
speza: @Skarlso are you doing these manually or using a formatter?

Skarlso:
@speza Quick question. Did we ever consider using [Casbin](https://casbin.org) with echo: https://echo.labstack.com/middleware/casbin-auth

Skarlso:
casbin and it's configration based role set + middleware actually looks kind of interesting.... 
https://casbin.org/docs/en/syntax-for-models

speza:
Hmm I do remember seeing this in the past. Not for Gaia though. I wonder if this seemed a little heavy weight... but arguably we’re heading towards this now.

handlers/handler.go
Skarlso: By hand, I'm afraid. 😁

security/rbac/enforcer.go
```diff
@@ -0,0 +1,154 @@
...
+	if !nsExists {
+		return errNamespaceNotFound
+	}
+
+	act, actionExists := ns[cfg.Action]
+	if !actionExists {
+		return errActionNotFound
+	}
+
+	if cfg.Resource == "" {
```
Skarlso, line 67: If the resource is empty it's all resources? I would think if the resource is empty, we shouldn't automatically allow all. `*` should be something very explicit. If new resources are defined, that means nothing is allowed.

security/rbac/enforcer.go
```diff
@@ -0,0 +1,154 @@
...
+	var stmts []gaia.RBACPolicyStatementV1
+	for policyName := range user.Policies {
+		policyResource, _ := s.svc.GetPolicy(policyName)
+		stmts = append(stmts, policyResource.Statement...)
+	}
+
+	eval := make(gaia.RBACEvaluatedPermissions)
+
+	// Evaluate all the policies, creating a single map and point of reference fro the user.
+	// This is not particularly efficient O(n2), but we have to parse all namespaces and actions.
```
Skarlso, line 104: *fro -> for
`O(n2)` -> `O(n^2)`

security/rbac/enforcer.go
speza: Yep, this was just temporary. I hadn’t got to implementing proper resource authentication yet mate 👍🏻 Was just to unblock testing on!

handlers/handler.go
```diff
@@ -24,41 +28,56 @@ func (s *GaiaHandler) InitHandlers(e *echo.Echo) error {
...
-		e.GET(p+"pipeline/name", pipelineProvider.PipelineNameAvailable)
-		e.POST(p+"pipeline/githook", GitWebHook)
-		e.GET(p+"pipeline/created", pipelineProvider.CreatePipelineGetAll)
-		e.GET(p+"pipeline", pipelineProvider.PipelineGetAll)
-		e.GET(p+"pipeline/:pipelineid", pipelineProvider.PipelineGet)
-		e.PUT(p+"pipeline/:pipelineid", pipelineProvider.PipelineUpdate)
-		e.DELETE(p+"pipeline/:pipelineid", pipelineProvider.PipelineDelete)
-		e.POST(p+"pipeline/:pipelineid/start", pipelineProvider.PipelineStart)
+
+		e.POST(p+"pipeline", pipelineProvider.CreatePipeline, policyEnforcer.do(rbac.PipelineNamespace, rbac.CreateAction))
```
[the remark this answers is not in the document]
speza: Let me have a think mate. I did something similar, albeit structs instead of YAML with the old system.

handlers/rbacmiddleware.go
```diff
@@ -0,0 +1,44 @@
...
+func newPolicyEnforcerMiddleware(enforcer rbac.PolicyEnforcer) *policyEnforcerMiddleware {
+	return &policyEnforcerMiddleware{
+		enforcer: enforcer,
+	}
+}
+
+func (pe *policyEnforcerMiddleware) do(namespace gaia.RBACPolicyNamespace, action gaia.RBACPolicyAction) echo.MiddlewareFunc {
+	return func(next echo.HandlerFunc) echo.HandlerFunc {
+		return func(c echo.Context) error {
+			if ctx, ok := c.(AuthContext); ok {
```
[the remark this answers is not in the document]
speza: Completely agree. I would usually do this 👍🏻

security/rbac/cache.go
```diff
@@ -0,0 +1,99 @@
...
+		for {
+			select {
+			case <-ticker.C:
+				c.EvictExpired()
+			}
+		}
+	}()
+
+	return c
+}
```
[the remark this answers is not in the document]
speza: Will take a look. Still getting my head around channels and goroutines!

security/rbac/cache.go
```diff
@@ -0,0 +1,99 @@
...
+
+// Put creates or updates an item. This uses the default expiration time.
+func (c *cache) Put(key string, value gaia.RBACEvaluatedPermissions) gaia.RBACEvaluatedPermissions {
+	defer c.mu.Unlock()
+	c.mu.Lock()
```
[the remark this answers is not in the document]
speza: Agree here. I think this was just a mistake on my part!

hsluoyz:
> Hmm @Skarlso. I do remember seeing this in the past. Not for Gaia though. I wonder if this seemed a little heavy weight... but arguably we’re heading towards this now.

@Skarlso @speza Hi guys! I'm Casbin author. I looked through the committed files. And it seems Casbin has already got all functionalities you need. Casbin only has 1 dependency (the evaluator) so it isn't very "heavy" in my sense. Of course, Casbin's code is complex than this PR's code but its correctness is proved by hundreds of test cases in the CI and production use of many companies. But your customized authz code may contain bugs and it's not easy to find them all via code review only. If one bug is not found here, maybe it will be a severe security vulnerability. If we can switch to using Casbin, we are very happy to solve your issues using it and then you can concentrate on your main directions instead of authz.

Skarlso:
@hsluoyz Hi!

Absolutely awesome from you to reach out to us. :)

Let us formulate an answer for you and we'll see if you can help out with that or maybe we misunderstood something about Casbin that you could clarify. :)

speza:
Yeah thanks for reaching out @hsluoyz! 

I will try and remember my reasons for thinking Casbin was heavy. I definitely think the code I have written here isn’t particularly simple anymore to be honest, and also 100% not a fan of not reinventing the wheel that’s for sure.

PR #250, opened 2020-05-26
Fix username to git if it is empty. Will document it

Skarlso:


workers/pipeline/git.go
```diff
@@ -310,7 +310,11 @@ func getAuthInfo(repo *gaia.GitRepo, callBack gossh.HostKeyCallback) (transport.
 		}
 	} else if repo.PrivateKey.Key != "" {
 		var err error
-		auth, err = ssh.NewPublicKeys(repo.PrivateKey.Username, []byte(repo.PrivateKey.Key), repo.PrivateKey.Password)
+		username := repo.PrivateKey.Username
```
speza, line 313: Is it worth a comment here to explain why we do this?

Skarlso: I wanted to open a PR in docs about this, but it's worth adding one in here as well I suppose.

PR #251, opened 2020-05-26
Fixed trigger check

Skarlso:


handlers/auth.go
```diff
@@ -18,19 +18,23 @@ var (
 	errNotAuthorized = errors.New("no or invalid jwt token provided. You are not authorized")
 
 	// Non-protected URL paths which are prefix checked
-	nonProtectedPathsPrefix = []string {
+	nonProtectedPathsPrefix = []string{
```
speza, line 21: Could this maybe be a single list that uses regex instead?
We could potentially combine all three lists then?

Skarlso: Regex is slow, that's why we avoid it in this place which runs fairly often.

speza: Sure. Although I don't think these slices would ever be long enough to notice. But it works, I guess there is more important things to focus on! :)

speza: Actually, another thought...
Could we not use echo router groups instead? Just apply the middleware to a group instead?
https://echo.labstack.com/guide/routing

Skarlso: They are in different groups under different sub paths which would result in a tangle of groups in various segments. The pipeline group has the trigger and the githook outside of that group, but still inside for other purposes. The js and the rest don't have a group at all.

speza: Maybe I'm missing something here dude, but this seems much nicer to me - https://github.com/gaia-pipeline/gaia/compare/master...speza:echo-groups

Skarlso: I'm all about adding that in a separate PR once it's tested nicely that all endpoints work as expected. :)

PR #255, opened 2020-06-07
Started to update docker images

Skarlso:


Skarlso:
I need to do some testing with all the new versions. Also, I'm going to deploy a test gaia instance which will run constantly so updates and upgrades can happen with long living pipelines.

Skarlso:
I've created a long living deployment, which I'm going to update with these. I'll create a pipeline with each of these with the old version, then update, and every pipeline should still work nicely.

Skarlso:
```
cppsdk/plugin.pb.h:17:2: error: #error This file was generated by an older version of protoc which is
#error This file was generated by an older version of protoc which is
```

Seems like I can't update protoc unless I update cppsdk as well.

Skarlso:
Also node... same for the nodejs sdk. So I leave that out for now to minimise the changes.

Skarlso:
All pipelines have been reviews and tested.

PR #256, opened 2020-06-08
RBAC using Casbin

speza:
WIP


speza:
Ok, so this is getting there now. This change introduces Casbin to enforce RBAC.

We have a policy model that is defined like so (see static/rbac-model.conf).
p, role:admin, pipelines, get, *, allow

Namespace = pipelines, secrets, etc.
Action = create, get, delete, etc
Resource = pipelineid, etc

We use manage a standard set of policy rules loaded in on server start ( static/rbac-policy.csv).

**Summary:**
- Implemented an API mapping yaml file that stores all our endpoints and stores which namespace/action/resource to apply when we enforce RBAC
- We now instantiate a casbin adapter that sits on top of boltdb and saves each policy rule line as a k/v pair
- Use rice to include some more static assets (rbac-model.conf, rbac-policy.csv, rbac-api-mappings.yml
- Implemented some new endpoints to manage rbac

// RBAC - Management
apiAuthGrp.GET("rbac/roles", rbacHandler.getAllRoles)
apiAuthGrp.PUT("rbac/roles/:role", rbacHandler.addRole)
apiAuthGrp.DELETE("rbac/roles/:role", rbacHandler.deleteRole)
apiAuthGrp.PUT("rbac/roles/:role/attach/:username", rbacHandler.attachRole)
apiAuthGrp.DELETE("rbac/roles/:role/attach/:username", rbacHandler.detatchRole)
apiAuthGrp.GET("rbac/roles/:role/attached", rbacHandler.getRolesAttachedUsers)
// RBAC - Users
apiAuthGrp.GET("users/:username/rbac/roles", rbacHandler.getUserAttachedRoles)

- Add more logic to the authMiddleware to enforce RBAC

**TODO:**
- Add the ability to turn RBAC on/off
- More testing
- Another PR to implement the UI

static/rbac-policy.csv
```diff
@@ -0,0 +1,36 @@
+# Built-in policy which defines two roles - role:admin and role:readonly.
+# role:admin is automatically added to the admin user.
+# Policy format:
+# p, <role>, <namespace>, <action>, <resource>, <effect = allow/deny>
+
+p, role:admin, pipelines, get, *, allow
```
hsluoyz, line 6: If `role:admin` is assumed to access all, you can just write `p, role:admin, *, *, *, allow` one rule to allow all.

security/rbac/endpoint_enforcer_test.go
```diff
@@ -0,0 +1,111 @@
+package rbac
+
+import (
+	"errors"
+	"github.com/casbin/casbin/v2"
```
hsluoyz, line 5: Sort the imports.

security/rbac/service.go
```diff
@@ -0,0 +1,157 @@
...
+	return &enforcerService{
+		enforcer:        enforcer,
+		rbacapiMappings: rbacapiMappings,
+	}, nil
+}
+
+func loadModel() (model.Model, error) {
+	modelStr, err := assethelper.LoadRBACModel()
+	if err != nil {
+		return nil, err
```
Skarlso, line 71: Add logging here as well, and all the rest of the errors. The more messages we get the more we know what went wrong. Pass in the logger as a dependency so you can use it with a method receiver.

security/rbac/service.go
```diff
@@ -0,0 +1,157 @@
...
+	return model, nil
+}
+
+func loadAPIMappings() (gaia.RBACAPIMappings, error) {
+	mappings, err := assethelper.LoadRBACAPIMappings()
+	if err != nil {
+		return gaia.RBACAPIMappings{}, err
+	}
+
+	var rbacapiMappings gaia.RBACAPIMappings
```
Skarlso, line 88: Declare this above and then return it.

security/rbac/service.go
```diff
@@ -0,0 +1,157 @@
...
+	if err := yaml.Unmarshal([]byte(mappings), &rbacapiMappings); err != nil {
```
Skarlso, line 89: I wonder if we should lock this mapping down for tempering with UnmarshalStrict. That will throw errors in case of duplicate fields and mappings which do not have corresponding structs.

security/rbac/service.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// DetachRole detaches a role from a user.
+func (e *enforcerService) DetachRole(username string, role string) error {
+	exists, err := e.enforcer.DeleteRoleForUser(username, role)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		return errors.New("role does not exists for user")
```
Skarlso, line 154: *exist

helper/assethelper/helper.go
```diff
@@ -0,0 +1,32 @@
+package assethelper
+
+import (
+	rice "github.com/GeertJohan/go.rice"
+)
+
+func loadStaticFile(filename string) (string, error) {
+	box, err := rice.FindBox("../../static")
```
Skarlso, line 8: I wonder if we should test this on a different OS working as well.. Probably is, just wondering.

helper/assethelper/helper.go
```diff
@@ -0,0 +1,32 @@
...
+	filestr, err := box.String(filename)
+	if err != nil {
+		return "", err
+	}
+	return filestr, nil
+}
+
+// LoadRBACBuiltinPolicy loads the builtin rbac-policy.csv
+func LoadRBACBuiltinPolicy() (string, error) {
+	return loadStaticFile("rbac-policy.csv")
```
Skarlso, line 21: Put the file names into constants.

store/store.go
```diff
@@ -119,6 +125,15 @@ func (s *BoltStore) Init(dataPath string) error {
 	}
 	s.db = db
 
+	// TODO: Having Casbin stuff here doesn't sit quite right with me (especially loading a file here).
+	// Unfortunately we need to re-use the open bolt database for the adapter though.
+	builtinPolicy, _ := assethelper.LoadRBACBuiltinPolicy()
+	casbinAdapter, err := boltadapter.NewAdapter(db, "casbin-policies", builtinPolicy)
```
Skarlso, line 131: create a casbin store provider. The New should get the db and the policy and the rest of the parameters.
Then in server.go where we initialise the store, you can initialise the casbinProvider. Then pass that provider in where ever we might use / need it. The NewCasbinProvider then can load the file and return an error in case there is a problem which would/should essentially halt starting Gaia since the casbin stuff is part of it even if it is disabled in configs later on.

Makes sense?

Skarlso (commented):
Overall I agree with the approach. Well done. :) Just have a few remarks regarding the code itself and a tiny bit of structure in storage.

security/rbac/endpoint_enforcer.go
```diff
@@ -0,0 +1,50 @@
...
+		return true, nil
+	}
+
+	perm, ok := endpoint.Methods[method]
+	if !ok {
+		gaia.Cfg.Logger.Warn("method not mapped to api group path", "path", path, "method", method)
+		return true, nil
+	}
+
+	splitAction := strings.Split(perm, "/")
```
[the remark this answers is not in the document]
speza: Hm that's an interesting thought. We should probably handle it you're right.
Unless we name it something like `worker::pipeline/create`? Actually I think `worker/pipeline::create` maybe even looks nicer? 

My main thoughts are that we should seperate the `namespace` and `action` with something unique (in this case `/`).

security/rbac/service.go
speza: WIll go through and do this as I tidy up mate

security/rbac/endpoint_enforcer.go
```diff
@@ -0,0 +1,50 @@
...
+	namespace := splitAction[0]
+	action := splitAction[1]
+
+	fullResource := "*"
```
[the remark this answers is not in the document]
speza: So we need to use `*` for endpoints where there isn't a resource id specified in the api-mappings yaml. Right now that is something like `secrets/create` that doesn't need the enforcement of a resource id on it.

security/rbac/service.go
speza: I'd prefer to add logging at the caller level. So in this case it would be where we call `NewEnforcerSvc`. The rest of the logging I will add into the handler code, etc.

One thing I will do is wrap the errors though to give more context :+1:

Skarlso: That's good as well. High level logging is fine but that logging wraps at least 3 or more different errors. If even one of them is similar we'll have no idea where the error happened. Wrapping is a good solution to that. Love the new error system. 😊

speza: Where’s a good old Java stacktrace when you need it! Haha

Skarlso: Bwuhahaha, pssss, not so loud.. The other gophers might hear you. 😂😂😂

store/store.go
speza: Hmm. I'm not sure if this would work @Skarlso? We don't have access to the open bolt DB outside of gaia.Store BoltStore. So we wouldn't be able to instantiate a casbin provider inside `server.go` where we depend on the open bolt db.

speza: We could move the above code into `service_provider.go` and instantiate the casbin adapter there. Makes more sense. If this error'd the server would fail to launch since we call `store, err := services.StorageService()` in `server.go`.

security/rbac/endpoint_enforcer.go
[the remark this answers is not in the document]
Skarlso: A single colon is fine actually, I think. No need to bring PHP into this. :D

security/rbac/endpoint_enforcer.go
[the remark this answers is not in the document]
Skarlso: Wait, why isn't there a secrets and why doesn't that need enforcement? :D

store/store.go
Skarlso: The db is a dependency for the provider. You would create the casbin stuff after we created and returned bolt db.

security/rbac/endpoint_enforcer.go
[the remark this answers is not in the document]
speza: PHP.... Java... anything else? :P

security/rbac/endpoint_enforcer.go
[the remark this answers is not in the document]
speza: I think you have misunderstood mate! We have permissions on secrets. I just mean that `secrets/create` doesn't have resource (a key ID) enforced because you haven't created something yet!

So:
`not specified` < cannot create secrets
`p, secrets, create, *, deny` < cannot create secrets
`p, secrets, create, *, allow` < can create secrets

store/store.go
speza: Maybe I'm not following @Skarlso. I don't think the code is setup to do this without some rework. If we're instantiating a `CasbinStoreProvider` within `server.go` we need access to `db *bbolt.DB` there so it can be passed in. That object is privately held within the store logic. We only expose the store interface to `server.go` (rightly so). I don't see a nice way to do this since we create the bolt connection inside `init` rather than passing it in using dependency injection.

Skarlso: The casbin store needs to be a separate entity from the gaia store. It does use gaia store though. So implement the two things it needs ( if I see this correctly, there are only 2? ). Extend the Gaia store interface to include policy saving and deleting. Then pass in gaia store as a dependency to casbin store which is using gaia store in the back. For testing casbin we need to be able to inject in a memory based store anyways.

speza: No, so the Casbin adapter has multiple methods for read/write (see here - https://github.com/speza/casbin-bolt-adapter/blob/master/adapter.go). It doesn’t use the Gaia store implementation specifically - it only requires the open Bolt connection. I figured we wanted to use the one we already create/open. Once you give it an open bolt connect (and bucket name) the adapter completely takes over.

Let me have a deeper think on this one later mate 😊

speza:
Just to top it off @Skarlso circle is still using Go 1.12 so won’t compile the error wrapping!

store/store.go
Skarlso: Oh I see what you mean. Hmmm... Yeah two open connections are not going to work. Bolt doesn't like that. In fact it will throw an error if you try. Alright, we'll think of something. 😊

Skarlso:
https://github.com/gaia-pipeline/gaia/pull/256#issuecomment-643323772

Yes. Right now circleci is failing until we can update it and push a new image.

security/rbac/noop.go
```diff
@@ -0,0 +1,53 @@
+package rbac
+
+import "errors"
+
+var errNotEnabled = errors.New("rbac is not enabled")
```
Skarlso, line 5: @speza Shouldn't a no-op RBAC return nil instead? Wouldn't this mean that normal operations which require RBAC fail rather than being ignored as the OFF switch suggests?

store/store.go
```diff
@@ -51,12 +55,22 @@ const (
...
 // BoltStore represents the access type for store
 type BoltStore struct {
-	db *bolt.DB
+	db            *bolt.DB
+	casbinAdapter persist.BatchAdapter
+}
+
+// SettingsStore is the interface that defines methods needed to save settings config into the store.
+type SettingsStore interface {
+	SettingsPut(config *gaia.StoreConfig) error
```
Skarlso, line 64: why can't an RBACPut be a SettingsPut? Why are they two different functions? An interface shouldn't be explicit. It should be generic mostly. If there is a settingsRbacPut that should most likely be either a more specialised interface or a concrete implementation of settings put.

I see that it's putting / getting an RBACConfig instead of a StoreConfig, but that also means that it can't be a SettingsStore. 

What I mean is, either make rbac use settingsput or have a separate interface. :)

server/server.go
```diff
@@ -333,3 +345,29 @@ func findExecutablePath() (string, error) {
...
+			gaia.Cfg.Logger.Error("failed to create new enforcer", "error", err.Error())
+			return nil, err
+		}
+		gaia.Cfg.Logger.Info("rbac enabled")
+		return svc, nil
+	}
+
+	gaia.Cfg.Logger.Info("rbac disabled")
+	return rbac.NewNoOpService(), nil
+}
```
Skarlso, line 373:
```suggestion
func initRBACService(store store.GaiaStore) (rbac.Service, error) {
	// globally disabled, can still be enabled by settings.
	if !gaia.Cfg.RBACEnabled {
		settingsEnabled, err := store.SettingsRBACGet()
		if err != nil {
			gaia.Cfg.Logger.Error("failed to get rbac settings", "error", err.Error())
			return nil, err
		}
		// completely disabled.
		if !settingsEnabled.Enabled {
			gaia.Cfg.Logger.Info("rbac disabled")
			return rbac.NewNoOpService(), nil
		}
	}

	svc, err := rbac.NewEnforcerSvc(gaia.Cfg.RBACDebug, store.CasbinStore())
	if err != nil {
		gaia.Cfg.Logger.Error("failed to create new enforcer", "error", err.Error())
		return nil, err
	}
	gaia.Cfg.Logger.Info("rbac enabled")
	return svc, nil
}
```

security/rbac/noop.go
speza: @Skarlso my thoughts here are that we should return an error for any explicit RBAC operations. E.g. trying to add a role to a user, remove a role, etc - purely so those actions don't display as successful (that would be misleading). I'd rather us inform the user that RBAC is not enabled. 

For `Enforce` its slightly different, we return no error so enforcement is successful since we don't want to block endpoints when RBAC isn't on.

Skarlso:
Testing was successful on the deployed instance! Well done @speza! Thanks for this awesome work and effort you've put into this and your perseverance. :) Fantastic work! 👍

PR #257, opened 2020-06-09
Updated circleci image

Skarlso:


Skarlso:
This will fail the build for now, until a new version is used.

PR #271, opened 2020-07-07
Delete user rbac attachments on deletion of users

speza:
* Add a DeleteUser method so we can delete rbac attachments for a user.
* Make noop.go fully no-op (makes sure that a call to DeleteUser) even with RBAC off is OK. All other endpoints now fully no-op for consistency.
* Introduce a UserHandler to contain dependencies for the user handlers


speza:
> This looks fine but please extract the user handler creation into server.go so init handler remains testable. InitHandler just used the user handler and shouldn't create it. It should be injected like the rest of the handlers.

@Skarlso do you want me to make a UserProivder? and I did the same with RBAC, so could make an RBACProvider?

Skarlso:
LGTM will merge after testing on the test instance.

Skarlso:
Will merge soon

PR #281, opened 2020-08-02
fix: rbac fixes and some refactoring

speza:
* Call GetAllSubjects() instead for listing RBAC roles. GetAllRoles() only gets roles that have actually been assigned to a user. This is because 'p' policy lines are defined for the roles. 'g' policy lines are for assignment of a role to a user. GetAllRoles() calls the 'g' line at index 2 to list roles.

* Fixed a bug with role:readonly looking for wildcard 'get' action which no longer exists because our actions became more specific - e.g. 'pipelines:runs, get-run' rather than 'pipelines:runs, get'

* Added a check to make sure all newly created roles have a prefix of 'role:x'

* Made the instantiation of the rbac more testable by using DI instead of instantiating concrete resources within the constructor

* Add missing service.go unit tests

### DIFF
--- a/security/rbac/service.go
+++ b/security/rbac/service.go
@@ -3,14 +3,21 @@ package rbac
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/casbin/casbin/v2"
 	"github.com/casbin/casbin/v2/model"
-	"github.com/casbin/casbin/v2/persist"
 	"gopkg.in/yaml.v2"
 
 	"github.com/gaia-pipeline/gaia/helper/assethelper"
 )
+
+// rolePrefix is the prefix we give to the policy lines in the Casbin model for a role.
+//  Roles are saved following this structure:
+//  p, role:myrole, *, get-thing, *, allow
+//  But individual user policies could look like:
+//  p, myuser, *, get-thing, *, allow
+const rolePrefix = "role:"
 
 type (
 	apiMapping struct {
@@ -23,7 +30,8 @@ type (
 		Resource string `json:"resource"`
 	}
 
-	apiLookup         map[string]apiLookupEndpoint
+	// APILookup is a map that can be used for quick lookup of the API endpoints that a secured using RBAC.
+	APILookup         map[string]apiLookupEndpoint
 	apiLookupEndpoint struct {
 		Param   string            `yaml:"param"`
 		Methods map[string]string `yaml:"methods"`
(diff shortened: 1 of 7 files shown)


speza:
@Skarlso some fixes for RBAC, more tests, and general quality of life changes.

Skarlso:
Oh f*ck off. :D @speza You have a `/` in your PR which is an invalid docker tag character. :D

speza:
@skarlso 🤣 just finding bugs in your code bro!

Skarlso:
yeah :D

Skarlso:
@gaia-bot OH FUCK OFF

speza:
@Skarlso would be cool if the bot posted the link to the running service... I'm lazy! 🤣

Skarlso:
https://gaia.cronohub.org :P It didn't change. :D

server/server.go
```diff
@@ -380,11 +383,21 @@ func initRBACService(store store.GaiaStore) (rbac.Service, error) {
 		}
 	}
 
-	svc, err := rbac.NewEnforcerSvc(gaia.Cfg.RBACDebug, store.CasbinStore())
+	model, err := rbac.LoadModel()
+	if err != nil {
+		return nil, fmt.Errorf("error loading model: %w", err)
```
Skarlso, line 388: Please Log the error and return a userfriendly message. The log is enough to check out the rest. And below too.

security/rbac/service.go
```diff
@@ -51,54 +59,36 @@ type (
...
+		rbacAPILookup: rbacAPILookup,
+	}
 }
 
-func loadModel() (model.Model, error) {
+// LoadModel loads the rbac model string from the assethelper and parses it into a Casbin model.Model.
+func LoadModel() (model.Model, error) {
 	modelStr, err := assethelper.LoadRBACModel()
 	if err != nil {
 		return nil, fmt.Errorf("error loading rbac model from assethelper: %w", err)
```
Skarlso, line 79: Here too, please return something user friendly and log the actual error.

speza: @Skarlso I prefer to wrap errors so we can more easily debug. Its up to the caller of this method to log. Therefore I'd prefer adding the logf into `server.go#initRBACService` (for which I definitely forgot!). If this specific error occurred it would be more for us to diagnose a problem than a user issue. So I'd rather include as much context as possible?

server/server.go
speza: We wrap the error and log in the calling code. I think these errors are only really for us anyway?
We could return a user friendly error in the calling code that is just a blanket "failed to init RBAC". But here, I'd prefer to wrap and then log where we call initRBACService.

security/rbac/service.go
Skarlso: In that case use `errors.Wrap`?

speza: `fmt.Errorf("error loading rbac model from assethelper: %w", err)` is wrapping errors? Its the Go std lib way of wrapping errors introduced in 1.13 :)

Skarlso: Ah right, I forgot. :D But then in the upper calls we'll need to do some logic in unwrapping?

speza: When you do the log on the error it will show the full error string of all errors. So something like `error a: error b: error c` all the way down the error stack). But other benefits are being able to use `errors.Is` and `errors.As` to look for any errors in the wrapped error.

Skarlso: Yes, that's why I asked that we would need to inspect the errors if it's something we can recover from. Anyways, fine then. I just don't want anything to creep up to the user. But I followed through that calls so this should be fine.

speza: :+1: Let me just make this code a little bit more consistent dude. You're right though, in some circumstances we'd definitely want to recover. I feel like here if something fails to load (like the model) we're buggered though! Recovery could be turning RBAC off, but that would be bad if it was meant to be on!

Skarlso: Yes, let's not do that. :) ( turning it off I mean. )

speza: Haha! Don't worry!

Skarlso:
Holy shit. `npm` is down.

speza:
> Holy shit. `npm` is down.

Hoping it was just a wobble... https://status.npmjs.org/

Issue #150, opened 2019-01-15
Document the API and using Gaia without the Admin client


Skarlso:
There has been need to not use the UI but rather the API directly. We need to document a proper way of using the API without the admin client and document the API as well and it's features.

michelvocks:
@Skarlso thanks. We need this for sure!

I also think that the current way of triggering a Gaia pipeline is too complex and not really user friendly (send request to get bearer token, send another request to trigger pipeline).
It would be cool if (optional) Gaia would automatically generate a token (or user provides one) and this could be used to trigger one specific pipeline. What do you think?

Skarlso:
@michelvocks Hmm. How would you get to that token? Or would you have to use an authenticated admin for it which gives it to you for later usage? We'd have to safe the triggering token / pipeline, which should be fine. 🤔

michelvocks:
@Skarlso 

> Hmm. How would you get to that token? Or would you have to use an authenticated admin for it which gives it to you for later usage?

Exactly. In the pipeline settings you should be able to set/generate a token and attach it to one specific pipeline. 
I know this feature from jenkins: https://humanwhocodes.com/blog/2015/10/triggering-jenkins-builds-by-url/
But there it's in my opinion also too complicated. I think there is no reason for providing two tokens.

Skarlso:
Cool! That sounds doable!

leberknecht:
It would be great if you could provide open API / swagger specs for the API, so we can use existing tooling like swagger-codegen to generate clients for SDK :)

Skarlso:
![ill-consider-it](https://user-images.githubusercontent.com/182850/51311328-f6143e00-1a48-11e9-99ed-d0fd33e901a2.jpg)

Skarlso:
Swagger has been added. I'll update the gaia docs with a static rip off it. Otherwise, the docs are available under:
the deployed base url /swagger/index.html.

Skarlso:
Documentation and swagger has been added.

Better late then never, I guess @leberknecht :D

PR #287, opened 2020-08-11
Adding swagger

Skarlso:
Generate the docs:

```
swag init -g ./cmd/gaia/main.go
```

Then view:

https://gaia.whatever.com/api/v1/swagger/index.html


Skarlso:
This will now fix #150.

Skarlso:
@speza Doing this I found a little mix-up in the rbac endpoints. :) It looks like you were using the wrong handler for one of the endpoints also the naming and comments of some other handlers were wrong which I didn't spot either. :D I fixed them in this PR. :)

Skarlso:
Hey @speza.

Do we still use this? `apiAuthGrp.GET("permission", PermissionGetAll)`.

Skarlso:
Nice! This works perfectly fine. :)

speza:
Ah nice! It was so much manual work I can believe I got something wrong haha. 

That old permissions endpoint is still being used if anyone used legacy perms. Still used in the UI until we deprecate it.

